PR #3, opened 2025-01-10
Convert kueue operator to use library-go (following secondary scheduler)

kannon92:
Convert kueue operator to use library-go (following secondary scheduler)


cmd/main.go
```diff
@@ -1,124 +1,19 @@
...
-
-Unless required by applicable law or agreed to in writing, software
-distributed under the License is distributed on an "AS IS" BASIS,
-WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
-See the License for the specific language governing permissions and
-limitations under the License.
-*/
-
-package main
+package operator
```
haircommander, line 1: it's possible we need to keep this

kannon92: good call!

PR #20, opened 2025-01-31
remove pod based webhooks from kueue webhooks

kannon92:
If there are no pod based frameworks, we should remove these from the webhook configuration.


pkg/builders/webhook/pod_webhook.go
```diff
@@ -0,0 +1,83 @@
...
+	}
+	newWebHook := currentWebhook.DeepCopy()
+	newWebHook.Webhooks = []admissionregistrationv1.ValidatingWebhook{}
+	webhooksToSearchFor := dangerousWebhooks()
+
+	for _, val := range currentWebhook.Webhooks {
+		removeWebhook := false
+		for _, name := range webhooksToSearchFor {
+			if val.Name == name {
+				removeWebhook = true
```
rphillips, line 47:
```suggestion
				removeWebhook = true
				break
```

pkg/builders/webhook/pod_webhook.go
```diff
@@ -0,0 +1,83 @@
...
+			}
+		}
+		if !removeWebhook {
+			newWebHook.Webhooks = append(newWebHook.Webhooks, val)
+		}
+	}
```
rphillips, line 81: This could be a function. It's used twice.

pkg/builders/webhook/pod_webhook.go
```diff
@@ -0,0 +1,83 @@
...
+
+import (
+	admissionregistrationv1 "k8s.io/api/admissionregistration/v1"
+
+	kueue "github.com/openshift/kueue-operator/pkg/apis/kueueoperator/v1alpha1"
+)
+
+// These are webhooks that we only want to enable if the integration requests it
+// Otherwise we will remove them from the kueue manifest
+func dangerousWebhooks() []string {
```
rphillips, line 27: This could be a static array and not a function.

mrunalp: Rename this to `optionalWebhooks` or something different. We do intend to deploy them on specific namespaces :)

pkg/builders/webhook/pod_webhook.go
kannon92: updated.

pkg/builders/webhook/pod_webhook.go
kannon92: updated.

rphillips: Is there a reason why a function is preferred here? A global static array would produce less GC garbage.

kannon92: I wasn't sure how to declare a constant static array.

kannon92: updated.

sohankunkerkar (approved):
/lgtm
Looks sane to me!

PR #22, opened 2025-01-31
bindata: add sync manifests script

sohankunkerkar:


Makefile
```diff
@@ -150,3 +150,14 @@ OPERATOR_SDK = $(shell which operator-sdk)
 endif
 endif
 
+.PHONY: sync-manifests
+sync-manifests:
```
kannon92, line 154: I've been bitten by rolling python3 code before. Can we use a container for podman so that we can control the actuall version that is used.

This can be a followup.

hack/sync_manifests.py
```diff
@@ -0,0 +1,184 @@
...
+        return dumper.represent_scalar("tag:yaml.org,2002:str", data, style="|")
+    # Check if the string was originally quoted with double quotes
+    if hasattr(data, "style") and data.style == '"':
+        return dumper.represent_scalar("tag:yaml.org,2002:str", data, style='"')
+    return dumper.represent_scalar("tag:yaml.org,2002:str", data)
+
+yaml.add_representer(str, string_representer)
+
+# Get the latest version of Kueue.
+api_url = "https://api.github.com/repos/kubernetes-sigs/kueue/releases/latest"
```
kannon92, line 24: Can we make this script take a version?

ie python3 hack/sync_manifests.py 0.10.1?

I just want to make sure that we have deterministic behavior on the script and not to just always take the latest version.

We may have to provide patches for 0.10 kueue or 0.12 so we probably want to be able to generate bindata based on the version of kueue we provide.

hack/sync_manifests.py
```diff
@@ -0,0 +1,184 @@
...
+
+    # Update namespace for specific resources (excluding Secret)
+    if kind in namespace_updates and "metadata" in doc:
+        doc["metadata"]["namespace"] = "openshift-kueue-operator"
+
+    # Preserve OpenShift-specific label for webhook-service.yaml
+    if kind == "Service" and doc["metadata"]["name"] == "kueue-webhook-service":
+        if "labels" not in doc["metadata"]:
+            doc["metadata"]["labels"] = {}
+        doc["metadata"]["labels"]["service.beta.openshift.io/serving-cert-secret-name"] = "kueue-webhook-server-cert"
```
sohankunkerkar, line 111: It looks like we don't need https://github.com/openshift/kueue-operator/pull/24/files#diff-64d97c2af009e5921341ae981a3cade926351feb0f38b4d7a24abd44b9703aadL8 anymore. I'm going to remove it from the script.

Makefile
sohankunkerkar: Fixed

Makefile
```diff
@@ -150,3 +150,18 @@ OPERATOR_SDK = $(shell which operator-sdk)
...
 
+
+# Use this target like make sync-manifests VERSION=<version>
+.PHONY: sync-manifests
+sync-manifests:
+	@echo "Syncing manifests in bindata/assets/kueue-operator using a container"
+	@podman run --rm \
+		-v $(PWD):/workspace:Z \
+		-w /workspace \
+		python:3.11-slim \
```
kannon92, line 161: If we put this in the CI, I think we would want to use a quay or red hat registry for this.

For local development this is fine.

PR #25, opened 2025-02-03
Misc updates

sohankunkerkar:


kannon92:
/lgtm

A PR template would also be useful but can be done as a followup.

PR #27, opened 2025-02-05
Add API fields based on feedback from RHOAI

kannon92:
/cc @KPostOffice @varshaprasad96


kannon92:
> The changes look good to me! Just one question - are we looking to enable feature gates too through Kueue CRs?

Yep! That is in the API. I added a change into upstream Kueue to allow setting those via configuration api.

config/schemapatch/kueue-operator.crd.yaml
```diff
@@ -193,6 +225,60 @@ spec:
...
+                    manageJobsWithoutQueueName:
+                      description: |-
+                        ManageJobsWithoutQueueName controls whether or not Kueue reconciles
+                        jobs that don't set the annotation kueue.x-k8s.io/queue-name.
+                        If set to true, then those jobs will be suspended and never started unless
+                        they are assigned a queue and eventually admitted. This also applies to
+                        jobs created before starting the kueue controller.
+                        Defaults to false; therefore, those jobs are not managed and if they are created
+                        unsuspended, they will start immediately.
+                      type: boolean
```
haircommander, line 237: probably should be a enum/string, booleans aren't considered good practice anymore

config/schemapatch/kueue-operator.crd.yaml
```diff
@@ -193,6 +225,60 @@ spec:
...
+                    managedJobsNamespaceSelector:
+                      description: ManagedJobsNamespaceSelector can be used to omit some namespaces from ManagedJobsWithoutQueueName
```
haircommander, line 239: technically isn't it the inverse: 
``
ManagedJobsNamespaceSelector can be used to include some namespaces for ManagedJobsWithoutQueueName
```

config/schemapatch/kueue-operator.crd.yaml
kannon92: Yea, but the config option is a bool. So if I add an enum I can convert it to a bool for the kueue config that kueue reads. But that seems a bit silly.

config/schemapatch/kueue-operator.crd.yaml
kannon92: https://github.com/kubernetes-sigs/kueue/blob/main/apis/config/v1beta1/configuration_types.go#L42

No, this is correct.

managedJobsWithoutQueueName would add a label to all jobs to automatically add them into kueue.

And then ManagedJobsNamespaceSelector was added as that proved a bit too strong of a design.

config/schemapatch/kueue-operator.crd.yaml
haircommander: I agree its silly but I think the point still stands

kannon92: so I will create a enum of True and False and convert that to a bool?

kannon92: Alright I went ahead and updated this.

PR #28, opened 2025-02-06
fix webhook bug where deployment, pod and statefulset were not dropped

kannon92:
Fix error where deployment, pod and statefulsets were still being used.


cpmeadors:
/lgtm

Tested on 4.17 AWS with "make deploy-ocp"  and I don't see the related error any more.

PR #31, opened 2025-02-10
OCPNODE-2916: add prometheus service monitor

kannon92:


hack/deploy-ocp.sh
```diff
@@ -9,6 +9,8 @@ hack/replace-image.sh deploy/examples KUEUE_IMAGE $KUEUE_IMAGE_REPLACE
 # Deploy the operator
 oc apply -f deploy/
 oc apply -f deploy/crd/
+echo "installed crd; wait for api to register then"
```
haircommander, line 12: nit them*

hack/deploy-ocp.sh
```diff
@@ -9,6 +9,8 @@ hack/replace-image.sh deploy/examples KUEUE_IMAGE $KUEUE_IMAGE_REPLACE
...
+sleep 10
```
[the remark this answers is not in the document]
kannon92: Sometimes I see that the crd api doesn't yet registry and then the example creation fails because the CRD was not yet defined.

It isn't used in tests but for mostly developers.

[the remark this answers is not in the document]
rphillips: could remove the sleep and the examples/job.yaml, then print the command to create the example.

pkg/operator/target_config_reconciler.go
```diff
@@ -791,6 +826,45 @@ func (c *TargetConfigReconciler) manageCertificateWebhookCR(ctx context.Context,
...
+			},
+			Endpoints: []monitoringv1.Endpoint{
+				{
+					Interval:        "30s",
+					Port:            "metrics", // Name of the port you want to monitor
+					Scheme:          "https",
+					BearerTokenFile: "/var/run/secrets/kubernetes.io/serviceaccount/token",
+					TLSConfig: &monitoringv1.TLSConfig{
+						SafeTLSConfig: monitoringv1.SafeTLSConfig{
+							InsecureSkipVerify: ptr.To(true),
```
rphillips, line 855: Hmm. InsecureSkipVerify will skip the TLS verification.

pkg/operator/target_config_reconciler.go
```diff
@@ -223,6 +221,28 @@ func (c TargetConfigReconciler) sync() error {
 		specAnnotations["rolebindings/leader-election"] = resourceVersion
 	}
 
+	if roleBindings, _, err := c.manageRole(kueue, "assets/kueue-operator/role-prometheus.yaml"); err != nil {
+		klog.Error("unable to create role prometheus")
+		return err
+	} else {
+		resourceVersion := "0"
```
haircommander, line 228: what happens if it stays this version?

rphillips: I reviewed the manageRole() function and the underlying API call. I don't think we need the nil check after the resourceVersion, nor the `resourceVersion := 0` declaration at all.

hack/deploy-ocp.sh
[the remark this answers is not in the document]
kannon92: You bring up a good point @rphillips.

I use environment variables to sub the image in examples/jobs but I don't think we should expose Image on the CRD API.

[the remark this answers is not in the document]
kannon92: Opened up https://github.com/openshift/kueue-operator/pull/33 to fix this issue.

pkg/operator/target_config_reconciler.go
kannon92: I did a refactoring to address this everywhere. https://github.com/openshift/kueue-operator/pull/37

kannon92:
/hold

We should merge https://github.com/openshift/kueue-operator/pull/38 first to have clean history.

I brought that one in to test LocalQueueMetrics.

pkg/operator/target_config_reconciler.go
kannon92: We have more work to enable InsecureSkipVerify of false.

I posted an upstream issue [here](https://github.com/kubernetes-sigs/kueue/issues/4377).

I can create a followup ticket for this after the upstream work is merged and carried.

kannon92:
/hold

Still testing this.

kannon92:
/hold cancel

Tested this and everything is good again.

PR #33, opened 2025-02-12
remove image from kueue crd

kannon92:


pkg/operator/starter.go
```diff
@@ -79,6 +80,7 @@ func RunOperator(ctx context.Context, cc *controllercmd.ControllerContext) error
 		discoveryClient,
 		crdClient,
 		cc.EventRecorder,
+		os.Getenv("RELATED_IMAGE_OPERAND_IMAGE"),
```
haircommander, line 83: I'm probably missing something but it's not clear to me who is supposed to set this env. set in the pod spec of the operator?

test/e2e/bindata/assets/00_kueue-operator.crd.yaml
```diff
@@ -42,6 +42,39 @@ spec:
               config:
                 description: config that is persisted to a config map
                 properties:
+                  fairSharing:
```
haircommander, line 45: are these pieces directly related to this PR? they don't seem so to me

kannon92: They are not. But I realize that I didn't have any sync for the test data so the CRD rotted a little bit.

I can do this as a separate PR if you want. I added the sync when I generate the crd as part of this work.

pkg/operator/starter.go
kannon92: Correct.

test/e2e/bindata/assets/00_kueue-operator.crd.yaml
kannon92: I guess I should be better about commits then. Out of habit I always squash.

kannon92: Opened up https://github.com/openshift/kueue-operator/pull/34.

pkg/operator/starter.go
kannon92: Is this clear now?

haircommander: yup all good!

PR #37, opened 2025-02-14
Refactors of the main reconcile loop

kannon92:
I wanted to follow a similar pattern that lws-operator did.

1) Compute owner referneces once and pass that information into all objects that Kueue manages.
2) Webhooks must be before the deployment.
3) Namespaced resources will store the resource version in the deployment annotations.

We don't need that information for cluster wide resources.

### DIFF
--- a/pkg/operator/target_config_reconciler.go
+++ b/pkg/operator/target_config_reconciler.go
@@ -249,17 +249,17 @@ func (c TargetConfigReconciler) sync() error {
 	specAnnotations["service/webhook-service"] = resourceVersion
 
 	// From here, we will create our cluster wide resources.
-	if err := c.manageCustomResources(kueue, ownerReference); err != nil {
+	if err := c.manageCustomResources(ownerReference); err != nil {
 		klog.Error("unable to manage custom resource")
 		return err
 	}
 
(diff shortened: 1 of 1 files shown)


pkg/operator/target_config_reconciler.go
```diff
@@ -171,148 +178,119 @@ func (c TargetConfigReconciler) sync() error {
...
-	crdAnnotations, crdErr := c.manageCustomResources(kueue)
-
-	if crdErr != nil {
-		klog.Error("unable to manage custom resource")
-		return crdErr
 	}
-
-	for key, val := range crdAnnotations {
-		specAnnotations[key] = val
+	if cm != nil {
```
rphillips, line 186: cm is always non-nil, right?

kannon92: The code in library-go does return nil in some cases so I don't think we can make the assumption that it always be non-nil.

rphillips: Ok. It's probably because of this block where the err is not handled. We can keep the nil checks.

https://github.com/openshift/library-go/blob/c917e6d528b2d61b1702a36e9d614f9088c95f62/pkg/operator/resource/resourceapply/core.go#L344-L356

PR #38, opened 2025-02-14
add feature gates to config map so one can set them

kannon92:
Discovered a bug where feature gates are never populated to ConfigMap.

You could reproduce this bug:

```yaml
apiVersion: operator.openshift.io/v1alpha1
kind: Kueue
metadata:
  labels:
    app.kubernetes.io/name: kueue-operator
    app.kubernetes.io/managed-by: kustomize
  name: cluster
  namespace: openshift-kueue-operator
spec:
  config:
    featureGates:
      LocalQueueMetrics: true
    integrations:
      frameworks:
      - "batch/job" 
```

Without this code, Kueue will set feature gates based on their defaults. With this change, one is able to set different feature gates.


sohankunkerkar (approved):
/lgtm

PR #45, opened 2025-02-26
Operator should correctly cleanup operand resources when deleted

PannagaRao:
This PR improves the cleanup process by adding a finalizer to the Kueue instance.  The rationale behind adding finalizer to kueue instance rather than to the namespace is that since both the operator and operand pods reside in the same namespace, the namespace finalizer will not efficiently manage resource cleanup.

Implemented cleanup for:
Mutating & Validating Webhooks
Certificates and Issuers (in operator namespace)
CRs, ClusterRoles and ClusterRoleBindings

Attaching logs of the clean up for reviewers.
Screenshot From 2025-02-26 11-30-24
Screenshot From 2025-02-26 11-31-12
Screenshot From 2025-02-26 11-31-47


test/e2e/e2e_operator_test.go
```diff
@@ -238,6 +238,13 @@ func deployOperator() error {
 					return err
 				}
 				requiredSS := requiredObj.(*ssv1.Kueue)
+				if err := ssClient.KueueV1alpha1().Kueues(requiredSS.Namespace).Delete(ctx, requiredSS.Name, metav1.DeleteOptions{}); err != nil && !strings.Contains(err.Error(), "not found") {
```
kannon92, line 241: Can you add a e2e test to confirm this functionality all works?

deploy/02_clusterrole.yaml
```diff
@@ -142,4 +143,28 @@ rules:
       - patch
       - update
       - watch
+  - apiGroups:
```
kannon92, line 146: Why do we need operatorgroups in our API?

pkg/operator/target_config_reconciler.go
```diff
@@ -303,10 +326,282 @@ func (c TargetConfigReconciler) sync() error {
...
+						encounteredError = err
+					} else {
+						klog.Infof("Successfully deleted CR: %s/%s", cr.GetNamespace(), cr.GetName())
+					}
+				}
+			}
+		}
+	}
+
+	operatorGroupsToDelete := []string{"openshift-cert-manager-operator"}
```
kannon92, line 481: We should not delete cert manager operators.

pkg/operator/target_config_reconciler.go
```diff
@@ -162,10 +164,31 @@ func (c TargetConfigReconciler) sync() error {
 		UID:        kueue.UID,
 	}
 
+	if err := c.addFinalizerToKueueInstance(kueue); err != nil {
+		klog.Errorf("Failed to add finalizer to Kueue instance: %v", err)
```
rphillips, line 168: missing a return err

pkg/operator/target_config_reconciler.go
```diff
@@ -162,10 +164,31 @@ func (c TargetConfigReconciler) sync() error {
...
+
+		defer func() {
+			if err := c.removeFinalizerFromKueueInstance(kueue); err != nil {
+				klog.Errorf("Failed to remove finalizer from Kueue instance %s: %v", kueue.Name, err)
+			} else {
+				klog.Infof("Finalizer successfully removed from Kueue instance %s", kueue.Name)
+			}
+		}()
+
+		if cleanupErr := c.cleanUpResources(); cleanupErr != nil {
```
rphillips, line 186: Use `err` since it's cleaner. It's quicker to think about instead of thinking `cleanupErr`, what error is this?

pkg/operator/target_config_reconciler.go
```diff
@@ -303,10 +326,282 @@ func (c TargetConfigReconciler) sync() error {
 	if err != nil {
 		return err
 	}
+	return nil
+}
+
+const kueueFinalizer = "kueue.openshift.io/finalizer"
```
rphillips, line 332: Move to the top of the file.

pkg/operator/target_config_reconciler.go
```diff
@@ -303,10 +326,282 @@ func (c TargetConfigReconciler) sync() error {
...
 
+func (c *TargetConfigReconciler) addFinalizerToKueueInstance(kueue *kueuev1alpha1.Kueue) error {
+	if !controllerutil.ContainsFinalizer(kueue, kueueFinalizer) {
+		controllerutil.AddFinalizer(kueue, kueueFinalizer)
+		if _, err := c.operatorClient.Kueues(kueue.Namespace).Update(c.ctx, kueue, metav1.UpdateOptions{}); err != nil {
```
rphillips, line 337: Should use a patch construct. Patch will only update the fields that have changed.

pkg/operator/target_config_reconciler.go
```diff
@@ -303,10 +326,282 @@ func (c TargetConfigReconciler) sync() error {
...
 	return nil
 }
 
+func (c *TargetConfigReconciler) removeFinalizerFromKueueInstance(kueue *kueuev1alpha1.Kueue) error {
+	if !controllerutil.ContainsFinalizer(kueue, kueueFinalizer) {
+		return nil
+	}
+
+	err := retry.RetryOnConflict(retry.DefaultRetry, func() error {
+		latest, getErr := c.operatorClient.Kueues(kueue.Namespace).Get(c.ctx, kueue.Name, metav1.GetOptions{})
```
rphillips, line 352: use err

pkg/operator/target_config_reconciler.go
```diff
@@ -303,10 +326,282 @@ func (c TargetConfigReconciler) sync() error {
...
+		if getErr != nil {
+			return getErr
+		}
+		controllerutil.RemoveFinalizer(latest, kueueFinalizer)
```
rphillips, line 356: error check is missing here.

pkg/operator/target_config_reconciler.go
```diff
@@ -303,10 +326,282 @@ func (c TargetConfigReconciler) sync() error {
...
+		_, updateErr := c.operatorClient.Kueues(latest.Namespace).Update(c.ctx, latest, metav1.UpdateOptions{})
+		return updateErr
+	})
+
+	if err != nil {
+		klog.ErrorS(err, "Failed to remove finalizer from Kueue instance", "Kueue", kueue.Name)
+		return err
+	}
+	klog.Infof("Finalizer %s removed from Kueue instance %s", kueueFinalizer, kueue.Name)
+	// c.cleanUpOperatorRole()
```
rphillips, line 366: Commented code... Perhaps take it out?

pkg/operator/target_config_reconciler.go
```diff
@@ -303,10 +326,282 @@ func (c TargetConfigReconciler) sync() error {
...
+	return nil
+}
+
+func (c *TargetConfigReconciler) cleanUpResources() error {
+	ctx := context.TODO()
+	klog.Info("Pannaga deleted kueue cluster")
```
rphillips, line 372: Remove debug statements.

pkg/operator/target_config_reconciler.go
rphillips: Should pass a context to cleanUpResources... c.ctx could be used.

pkg/operator/target_config_reconciler.go
```diff
@@ -303,10 +326,282 @@ func (c TargetConfigReconciler) sync() error {
...
+	var encounteredError error
+
+	webhookTypes := []struct {
+		listFunc   func(context.Context, metav1.ListOptions) ([]string, error)
+		deleteFunc func(context.Context, string, metav1.DeleteOptions) error
+		name       string
+	}{
+		{
+			listFunc: func(ctx context.Context, opts metav1.ListOptions) ([]string, error) {
+				webhookList, err := c.kubeClient.AdmissionregistrationV1().MutatingWebhookConfigurations().List(ctx, opts)
```
rphillips, line 384: This enumerates all the webhooks in the cluster. Perhaps restrict it to just the namespaces we installed the Kueue webhook into?

pkg/operator/target_config_reconciler.go
```diff
@@ -303,10 +326,282 @@ func (c TargetConfigReconciler) sync() error {
...
+			} else {
+				klog.Infof("%s %s deleted successfully.", webhook.name, name)
+			}
+		}
+	}
+
+	crdList, err := c.crdClient.CustomResourceDefinitions().List(ctx, metav1.ListOptions{})
+	if err != nil {
+		klog.Errorf("Failed to list CRDs: %v", err)
+		encounteredError = err
```
rphillips, line 442: return err here

pkg/operator/target_config_reconciler.go
kannon92: Webhooks are cluster scoped. You can limit webhooks to only apply to certin namespaces but they are installed into the cluster.

pkg/operator/target_config_reconciler.go
```diff
@@ -303,10 +326,282 @@ func (c TargetConfigReconciler) sync() error {
...
+	// c.cleanUpOperatorRole()
+	return nil
+}
+
+func (c *TargetConfigReconciler) cleanUpResources() error {
+	ctx := context.TODO()
+	klog.Info("Pannaga deleted kueue cluster")
+	klog.Info("Starting cleanup of Kueue resources...")
+
+	var encounteredError error
```
rphillips, line 375: encounteredError has too many side effects in the function. Perhaps use  the [apimachinery](https://github.com/kubernetes/apimachinery/blob/v0.32.2/pkg/util/errors/errors.go)  package to aggregate a list of errors. You would use `[]error` to add an error to the list.

pkg/operator/target_config_reconciler.go
```diff
@@ -303,10 +326,282 @@ func (c TargetConfigReconciler) sync() error {
...
+		})
+		if err != nil {
+			klog.Errorf("Failed to delete Namespace %s: %v", ns, err)
+			encounteredError = err
+		} else {
+			klog.Infof("Successfully deleted Namespace: %s", ns)
+		}
+	}
+
+	clusterRoleList, err := c.kubeClient.RbacV1().ClusterRoles().List(ctx, metav1.ListOptions{})
```
rphillips, line 547: Restrict to namespace.

pkg/operator/target_config_reconciler.go
```diff
@@ -303,10 +326,282 @@ func (c TargetConfigReconciler) sync() error {
...
+	if err != nil {
+		klog.Errorf("Failed to list ClusterRoles: %v", err)
+		encounteredError = err
+	} else {
+		if len(clusterRoleList.Items) == 0 {
+			klog.Warning("No ClusterRoles found matching 'kueue'. Skipping deletion.")
```
rphillips, line 553: Probably superfluous.

test/e2e/e2e_operator_test.go
```diff
@@ -238,6 +238,13 @@ func deployOperator() error {
...
+					klog.Errorf("Failed to delete existing Kueue resource: %v", err)
+					return err
+				}
+
+				// Wait to ensure deletion is complete
+				time.Sleep(5 * time.Second)
```
rphillips, line 247: time.Sleep should not be needed. try using `wait.PollImmediate`.

pkg/operator/target_config_reconciler.go
PannagaRao: Deleting the cluster role related to kueue at the cluster scope so restricting to namespace wouldn't delete all of them

pkg/operator/target_config_reconciler.go
PannagaRao: Changed it to use Patch

deploy/02_clusterrole.yaml
PannagaRao: Removed it

pkg/operator/target_config_reconciler.go
```diff
@@ -303,10 +326,282 @@ func (c TargetConfigReconciler) sync() error {
...
+		}
+
+		for _, role := range clusterRoleList.Items {
```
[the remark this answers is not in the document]
PannagaRao: Split it into different functions.

pkg/operator/target_config_reconciler.go
```diff
@@ -324,7 +364,271 @@ func (c TargetConfigReconciler) sync() error {
...
+
+func (c *TargetConfigReconciler) addFinalizerToKueueInstance(kueue *kueuev1alpha1.Kueue) error {
+	if !controllerutil.ContainsFinalizer(kueue, KueueFinalizer) {
+		patch := []byte(`{"metadata":{"finalizers":["` + KueueFinalizer + `"],"resourceVersion":"` + kueue.ResourceVersion + `"}}`)
+		if _, err := c.operatorClient.Kueues(kueue.Namespace).Patch(c.ctx, kueue.Name, types.MergePatchType, patch, metav1.PatchOptions{}); err != nil {
+			klog.ErrorS(err, "Failed to add finalizer to Kueue instance", "Kueue", kueue.Name)
+			return err
+		}
+		klog.Infof("Finalizer %s added to Kueue instance %s", KueueFinalizer, kueue.Name)
+	}
```
rphillips, line 378:
```suggestion
        if controllerutil.ContainsFinalizer(kueue, KueueFinalizer) {
                return nil
        }
	patch := []byte(`{"metadata":{"finalizers":["` + KueueFinalizer + `"],"resourceVersion":"` + kueue.ResourceVersion + `"}}`)
	if _, err := c.operatorClient.Kueues(kueue.Namespace).Patch(c.ctx, kueue.Name, types.MergePatchType, patch, metav1.PatchOptions{}); err != nil {
		klog.ErrorS(err, "Failed to add finalizer to Kueue instance", "Kueue", kueue.Name)
		return err
	}
	klog.Infof("Finalizer %s added to Kueue instance %s", KueueFinalizer, kueue.Name)
	}
```

pkg/operator/target_config_reconciler.go
```diff
@@ -324,7 +364,271 @@ func (c TargetConfigReconciler) sync() error {
...
+
+func (c *TargetConfigReconciler) cleanUpCustomResources(ctx context.Context) error {
+	var errorList []error
+	crdList, err := c.crdClient.CustomResourceDefinitions().List(ctx, metav1.ListOptions{})
+	if err != nil {
+		klog.Errorf("Failed to list CRDs: %v", err)
+		return err
+	} else {
+		for _, crd := range crdList.Items {
+			if strings.Contains(crd.Name, "kueue") {
```
rphillips, line 483: Let's remove the nesting of the code by inverting the logic.

```
if !strings.Contains(crd.Name, "kueue") {
  continue
}
```

pkg/operator/target_config_reconciler.go
```diff
@@ -324,7 +364,271 @@ func (c TargetConfigReconciler) sync() error {
...
+					errorList = append(errorList, err)
+				} else {
+					klog.Infof("Successfully deleted CRD: %s", crd.Name)
+				}
+			}
+		}
+	}
 
+	certManagerCRDs := []string{
+		"certificates.cert-manager.io",
```
rphillips, line 550: We should not delete cert-manager.

pkg/operator/target_config_reconciler.go
```diff
@@ -324,7 +364,271 @@ func (c TargetConfigReconciler) sync() error {
...
+	return nil
+}
+
+func (c *TargetConfigReconciler) cleanUpClusterRoles(ctx context.Context) error {
+	var errorList []error
+	clusterRoleList, err := c.kubeClient.RbacV1().ClusterRoles().List(ctx, metav1.ListOptions{})
+	if err != nil {
+		klog.Errorf("Failed to list ClusterRoles: %v", err)
+		errorList = append(errorList, err)
+	} else {
```
rphillips, line 578: Else is superfluous here if there a return err.

pkg/operator/target_config_reconciler.go
```diff
@@ -324,7 +364,271 @@ func (c TargetConfigReconciler) sync() error {
...
+		for _, role := range clusterRoleList.Items {
+			if strings.Contains(role.Name, "kueue") && role.Name != "openshift-kueue-operator" {
```
rphillips, line 580: Invert the logic to continue so it removes a nest.

pkg/operator/target_config_reconciler.go
```diff
@@ -324,7 +364,271 @@ func (c TargetConfigReconciler) sync() error {
...
+	}
+	return nil
+}
+
+func (c *TargetConfigReconciler) cleanUpClusterRoleBindings(ctx context.Context) error {
+	var errorList []error
+	clusterRoleBindingList, err := c.kubeClient.RbacV1().ClusterRoleBindings().List(ctx, metav1.ListOptions{})
+	if err != nil {
+		klog.Errorf("Failed to list ClusterRoleBindings: %v", err)
+		errorList = append(errorList, err)
```
rphillips, line 608: just return err here and remove the else on line 609

pkg/operator/target_config_reconciler.go
kannon92: ClusterRoles are cluster scoped so there is no way to restrict this to a namespace.

test/e2e/e2e_operator_test.go
```diff
@@ -99,6 +99,65 @@ var _ = Describe("Kueue Operator", Ordered, func() {
...
+				}
+
+				validatingwebhooks, _ := kubeClientset.AdmissionregistrationV1().ValidatingWebhookConfigurations().List(ctx, metav1.ListOptions{})
+				for _, wh := range validatingwebhooks.Items {
+					if strings.Contains(wh.Name, "kueue") {
+						return fmt.Errorf("ValidatingWebhookConfiguration %s still exists", wh.Name)
+					}
+				}
+				return nil
+			}, operatorReadyTime, operatorPoll).Should(Succeed(), "Resources were not cleaned up properly")
```
kannon92, line 158: What about CRDs?

pkg/operator/target_config_reconciler.go
```diff
@@ -324,7 +364,281 @@ func (c TargetConfigReconciler) sync() error {
...
+		}
+	}
+
+	if len(errorList) > 0 {
+		return utilerror.NewAggregate(errorList)
+	}
+	return nil
+}
+
+func (c *TargetConfigReconciler) cleanUpCustomResources(ctx context.Context) error {
```
kannon92, line 476: Talking with @rphillips I'm not sure we want this behavior on default.

I guess for deleting the operator we can delete the CRDs but people may not want the CRDs purged.

kannon92: https://olm.operatorframework.io/docs/tasks/uninstall-operator/

pkg/operator/target_config_reconciler.go
kannon92: Correct. I think we should delete the namespaced values for certificates and issuers upon deletion of the operator.

But cert manager CRDs should be left alone.

test/e2e/bindata/assets/02_clusterrole.yaml
```diff
@@ -142,3 +142,23 @@ rules:
       - patch
       - update
       - watch
+  - apiGroups:
+      - kueue.x-k8s.io
```
kannon92, line 146: Why is this needed? I don't see us using this in the code at the moment.

PannagaRao: It is needed to list and deleted the CRDs

kannon92: I think this is used to delete the resources that a CRD defines. Your `manageCustomResources` I think deletes the actual CRD (ie `kubectl get crd` rather than `kubectl delete clusterqueue cq`.

I do not see any code that deletes the objects that kueue operand creates.

test/e2e/e2e_operator_test.go
PannagaRao: Added code to verify deletion of CRDs

PannagaRao:
This PR now handles the deletion of Webhooks (Mutating & Validating Webhooks), ClusterRoles & ClusterRoleBindings
Kueue Custom Resources and Certificates & Issuers (within the operator namespace)

The cleanup is triggered when the Kueue CR (named cluster) is deleted. PR does not remove the CRDs themselves, only the resources defined by them.

After the cleanup, the operator remains running and the sync() method continues reconciling resources.
- It recreates Certificates and Issuers
- It restores ClusterRoles, ClusterRoleBindings and the Webhooks
- Updates the ownerReferences to point to the UID of a newly created cluster if one is created again.

This reconciliation creates the deleted resources so the e2e test case originally written to verify the deletion can't be used. 

Is this expected behavior. Would appreciate your thoughts and suggestions. @rphillips @sohankunkerkar @kannon92

PannagaRao:
In order to ensure that reconciliation loop does not create resources which have a stale uid ( pointing to the deleted cluster) changes are made in the sync() to only reconcile the resources when there is no deletion timestamp set on the kueue cluster ( when there is an active kueue cluster CR )

This PR now cleans up the resources upon kueue cluster deletion and reconciles them only when there is a new cluster created.

Would appreciate your thoughts and suggestions. @rphillips @sohankunkerkar @kannon92

pkg/operator/target_config_reconciler.go
```diff
@@ -160,10 +165,50 @@ func (c TargetConfigReconciler) sync() error {
...
+
 	specAnnotations := map[string]string{
 		"kueueoperator.operator.openshift.io/cluster": strconv.FormatInt(kueue.Generation, 10),
 	}
 
+	if kueue.DeletionTimestamp != nil {
+		klog.Infof("Kueue instance %s is being deleted. Initiating cleanup...", kueue.Name)
+
+		defer func() {
+			if err := c.removeFinalizerFromKueueInstance(kueue); err != nil {
```
rphillips, line 181: If err is != nil, in this defer block, then the finalizer shouldn't be removed yet.

PannagaRao: Moving the finalizer deletion after the deletion of all other resources.

pkg/operator/target_config_reconciler.go
```diff
@@ -324,7 +369,278 @@ func (c TargetConfigReconciler) sync() error {
...
 		return err
 	}
+	return nil
+}
+
+func (c *TargetConfigReconciler) addFinalizerToKueueInstance(kueue *kueuev1alpha1.Kueue) error {
+	if controllerutil.ContainsFinalizer(kueue, KueueFinalizer) {
+		return nil
+	}
+	patch := []byte(`{"metadata":{"finalizers":["` + KueueFinalizer + `"],"resourceVersion":"` + kueue.ResourceVersion + `"}}`)
```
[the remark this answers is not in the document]
rphillips: I like the patch idea. It's much easier on the API server, and the patch is more likely to succeed than a full Update operation. I suggest the following helper:

```
strategicpatch.CreateTwoWayMergePatch(originalData, modifiedData, datastruct)
```

This snippet will create the patch object to use within the patch API call.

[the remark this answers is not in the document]
sohankunkerkar: Yeah, I like this idea. We can change `update` ->`patch` maybe? (not tested)

```go

     return retry.RetryOnConflict(retry.DefaultRetry, func() error {
        original, err := c.operatorClient.Kueues(kueue.Namespace).Get(c.ctx, kueue.Name, metav1.GetOptions{})
        if err != nil {
            return err
        }

        if add == controllerutil.ContainsFinalizer(original, KueueFinalizer) {
            return nil
        }

        modified := original.DeepCopy()
        mutator(modified, KueueFinalizer)

        originalData, err := json.Marshal(original)
        if err != nil {
            return fmt.Errorf("failed to marshal original object: %w", err)
        }

        modifiedData, err := json.Marshal(modified)
        if err != nil {
            return fmt.Errorf("failed to marshal modified object: %w", err)
        }

        patch, err := strategicpatch.CreateTwoWayMergePatch(
            originalData,
            modifiedData,
            kueuev1alpha1.Kueue{}, 
        )
        if err != nil {
            return fmt.Errorf("failed to create patch: %w", err)
        }

        patched, err := c.operatorClient.Kueues(original.Namespace).Patch(
            c.ctx,
            original.Name,
            types.StrategicMergePatchType,
            patch,
            metav1.PatchOptions{},
        )
        if err != nil {
            return err
        }

        *kueue = *patched
        klog.Infof("Finalizer %s %sed on Kueue %s", KueueFinalizer, finalizerOp, kueue.Name)
        return nil
    })
```

test/e2e/e2e_operator_test.go
```diff
@@ -99,8 +101,172 @@ var _ = Describe("Kueue Operator", Ordered, func() {
...
+				for _, resource := range certManagerResources {
+					gvr := schema.GroupVersionResource{
+						Group:    "cert-manager.io",
+						Version:  "v1",
+						Resource: resource,
+					}
+
+					klog.Infof("Verifying removal of %s instances", resource)
+
+					crList, err := dynamicClient.Resource(gvr).Namespace("").List(ctx, metav1.ListOptions{})
```
kannon92, line 202: If I understand this correctly you are looking at certmanager/issuers in the default namespace.

In this case I would expect that you would never actually get any resources. They are created in `namespace` of the operator not in "".

PannagaRao: You are right. I have restricted the namespace to operatorNamespace instead of using "" (all namespaces)

pkg/operator/target_config_reconciler.go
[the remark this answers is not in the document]
PannagaRao: Thanks for the suggestion code. Used it with minor tweaks.

pkg/operator/target_config_reconciler.go
```diff
@@ -324,7 +363,310 @@ func (c TargetConfigReconciler) sync() error {
...
+
+		gvr := schema.GroupVersionResource{
+			Group:    crd.Spec.Group,
+			Version:  crd.Spec.Versions[0].Name,
+			Resource: crd.Spec.Names.Plural,
+		}
+
+		klog.Infof("Fetching instances of CR type: %s", crd.Name)
+
+		crList, err := c.dynamicClient.Resource(gvr).Namespace("").List(ctx, metav1.ListOptions{})
```
kannon92, line 529:
```suggestion
		crList, err := c.dynamicClient.Resource(gvr).Namespace(c.operatorNamespace).List(ctx, metav1.ListOptions{})
```

test/e2e/e2e_operator_test.go
kannon92: I think you need one fix for CRD namespaces also.

PannagaRao: localqueues.kueue.x-k8s.io and workloads.kueue.x-k8s.io are namespace scoped and others are cluster scoped

kannon92: So what will we do if someone creates localqueues or workloads in namespace x?
DO we delete those or do we keep all namesacped resources?

PannagaRao: Made changes to restrict deletion of resources in localqueues and workloads CRDs  to the operatorNamespace. So the ones created in namespace x will not be deleted. 

Does this restriction make sense or do you suggest deletion all resources?

kannon92: I don't really understand what you are trying to do here.

PannagaRao: For CRDs that have scope == Namespaced (localqueues and workloads) this PR deletes the resources only in operatorNamespace. 
For CRDs that are cluster scoped this PR deletes all the resources under them when the kueue cluster is deleted.

kannon92: So users would never create localqueue and workloads in the operator namespace.

Localqueues are created in a separate namespace that a user would usually have access too. The operator namespace is usually limited to the resources needed to run the operator. And new APIs that the operator handles would be created in a new namespace.

PannagaRao: Removing the code to cleanup custom resources. PR now handles cleanup of webhooks, certificates & issuers, cluster roles and role bindings.

pkg/operator/target_config_reconciler.go
kannon92: @PannagaRao @rphillips can we resolve this discussion?

pkg/operator/target_config_reconciler.go
PannagaRao: Not cleaning up custom resources by default

test/e2e/bindata/assets/02_clusterrole.yaml
```diff
@@ -142,3 +142,23 @@ rules:
...
+    resources:
+      - admissionchecks
```
kannon92, line 148: This can be removed also right?

kannon92:
/hold

/lgtm

Putting a hold in case @sohankunkerkar or @rphillips want to give a final lgtm.

test/e2e/bindata/assets/02_clusterrole.yaml
kannon92: This isn't really hurting anything as its our test code so this can be a follow up.

pkg/operator/target_config_reconciler.go
PannagaRao: Didnt restrict it to operator-namespace since webhook is a cluster scope resource. Resolving this comment.

PR #67, opened 2025-03-06
[OCPNODE-2991] pkg/operator: add HA configuration for Kueue deployment

sohankunkerkar:


kannon92:
This is for the operator. We want HA for the operand really.

kannon92:
Operator is fine but the webhooks are related to the operand resources.

hack/sync_manifests.py
```diff
@@ -119,6 +119,32 @@ def clean_name(name, kind):
         for container in doc["spec"]["template"]["spec"]["containers"]:
             if container["name"] == "manager":
                 container["image"] = "${IMAGE}"
+        # --- Add HA Configuration ---
```
kannon92, line 122: Would it make more sense just to put this in the operator?

sohankunkerkar: I think that's what my previous commit was doing.

kannon92: The operator code. When we create the deployment.

sohankunkerkar: ah, I see your point now. Pragmatically.

kannon92:
Looking good.

I'm still not clear if we want multiple replicas for the operator deployment or not. @rphillips wdyt? 

There are no webhooks with the operator but should we have two replicas for that anyway?

rphillips:
HA for the operator can make sense in case that RHCOS node goes down for a reboot.

rphillips:
It looks like we could most of this logic into the main deployment.yaml file.

rphillips:
/lgtm cancel

Sohan is going to add leader election to true on this PR.

rphillips:
unit tests failed

sohankunkerkar:
Fixed the unit tests

PR #71, opened 2025-03-07
install via mod flag for ginkgo and controller-gen

kannon92:
This hopefully fixes o/release problem for running `make generate`.


kannon92:
/cc @sohankunkerkar @rphillips 

The coverity check seems to be getting oomkilled causing failures..

PR #75, opened 2025-03-10
bump up memory limits for sast-coverity check

kannon92:
ref: https://issues.redhat.com/browse/KFLUXSPRT-2167

Trying to increase memory limits to hopefully avoid OOMKills.


sohankunkerkar (approved):
/lgtm

PR #85, opened 2025-03-18
OCPNODE-3007: Do not install alpha crd by default

PannagaRao:

### DIFF
--- a/test/e2e/e2e_operator_test.go
+++ b/test/e2e/e2e_operator_test.go
@@ -34,6 +34,7 @@ import (
 	"github.com/openshift/library-go/pkg/operator/resource/resourceread"
 
 	corev1 "k8s.io/api/core/v1"
+	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	apiextv1 "k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset/typed/apiextensions/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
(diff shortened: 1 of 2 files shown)


sohankunkerkar (commented):
/approve

Code looks good. Can you add a Jira link/ description for this PR?

test/e2e/e2e_operator_test.go
```diff
@@ -99,6 +99,30 @@ var _ = Describe("Kueue Operator", Ordered, func() {
 				return fmt.Errorf("pod is not ready")
 			}, operatorReadyTime, operatorPoll).Should(Succeed(), "kueue pod failed to be ready")
 
+			By("Verifying that no v1alpha CRDs are installed")
```
kannon92, line 102: So by is more of a log message.

I think you may want to put this in a It block.

I think that naturally corresponds to a different test case.

test/e2e/e2e_operator_test.go
```diff
@@ -99,6 +99,30 @@ var _ = Describe("Kueue Operator", Ordered, func() {
...
+
+			Eventually(func() error {
+				ctx := context.TODO()
+				crdList, err := getApiExtensionKubeClient().CustomResourceDefinitions().List(ctx, metav1.ListOptions{})
```
kannon92, line 106: This would pass if there were no items in this list or there were no beta apis also.

Maybe let's verify that we the kueue apis.

PannagaRao:
This PR is to forbid the use of Alpha APIs in our API so the Kueue alpha APIs should not be installed as part of our operator.
https://issues.redhat.com/browse/OCPNODE-3007

test/e2e/e2e_operator_test.go
PannagaRao: Sure, will move it as a new test case

kannon92:
/retitle OCPNODE-3007: Do not install alpha crd by default

test/e2e/e2e_operator_test.go
PannagaRao: Added a check to verify kueue CRDs are installed

kannon92 (commented):
/lgtm

PR #72, opened 2025-03-07
WIP: Api review updates

kannon92:
Updating our API with the apireview.

PR #87, opened 2025-03-21
pkg: enable webhooks on opt-in namespaces

sohankunkerkar:


pkg/operator/target_config_reconciler.go
```diff
@@ -617,6 +619,19 @@ func (c *TargetConfigReconciler) cleanUpClusterRoleBindings(ctx context.Context)
 	return nil
 }
 
+func (c *TargetConfigReconciler) defaultKueueSpec(kueue *kueuev1alpha1.Kueue) {
+	if kueue.Spec.Config.Integrations.PodOptions.NamespaceSelector == nil {
+		kueue.Spec.Config.Integrations.PodOptions.NamespaceSelector = &metav1.LabelSelector{
```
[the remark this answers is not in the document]
kannon92: Though I don't think you need any types from Kueue here. You just need to define this and use it I think.

pkg/operator/target_config_reconciler.go
```diff
@@ -617,6 +619,19 @@ func (c *TargetConfigReconciler) cleanUpClusterRoleBindings(ctx context.Context)
 	return nil
 }
 
+func (c *TargetConfigReconciler) defaultKueueSpec(kueue *kueuev1alpha1.Kueue) {
+	if kueue.Spec.Config.ManagedJobsNamespaceSelector == nil {
+		kueue.Spec.Config.ManagedJobsNamespaceSelector = &metav1.LabelSelector{
+			MatchExpressions: []metav1.LabelSelectorRequirement{
+				{
+					Key:      "kueue.openshift.io/managed",
```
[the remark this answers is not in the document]
rphillips: The key should be global const somewhere.

sohankunkerkar:
Just a note:
API changes for the operator are still in flux, so we are no longer using `ManagedJobsNamespaceSelector` to set the label. I'll be working on top of [PR #72](https://github.com/openshift/kueue-operator/pull/72), where we’ll be hardcoding the labelSelector. I'll also be pushing the Pod integration changes along with e2e tests, which are working for me locally on an OCP cluster.

pkg/webhook/pod_webhook.go
```diff
@@ -17,62 +17,213 @@ limitations under the License.
...
+		"kubeflow.org/paddlejob":        "paddlejob",
+		"kubeflow.org/pytorchjob":       "pytorchjob",
+		"kubeflow.org/tfjob":            "tfjob",
+		"kubeflow.org/xgboostjob":       "xgboostjob",
+		"deployment":                    "deployment",
+		"statefulset":                   "statefulset",
+		"kueue.x-k8s.io/clusterqueue":   "clusterqueue",
+		"kueue.x-k8s.io/workload":       "workload",
+		"kueue.x-k8s.io/resourceflavor": "resourceflavor",
+		"kueue.x-k8s.io/cohort":         "cohort",
```
kannon92, line 46: I believe @PannagaRao work on not installing the alpha CRDs would mean that we should not have Cohorts yet.

Oh, we probably still have the webhooks..

sohankunkerkar: >Oh, we probably still have the webhooks..

exactly.

pkg/webhook/pod_webhook.go
```diff
@@ -17,62 +17,219 @@ limitations under the License.
...
+		enabledFrameworks[string(fw)] = true
 	}
-	newWebHook := currentWebhook.DeepCopy()
-	newWebHook.Webhooks = []admissionregistrationv1.ValidatingWebhook{}
 
-	for _, val := range currentWebhook.Webhooks {
-		if !findWebhook(val.Name, podBasedValidatingWebhooks) {
-			newWebHook.Webhooks = append(newWebHook.Webhooks, val)
+	for _, wh := range currentWebhook.Webhooks {
+		framework := getFrameworkForWebhook(wh.Name, "v")
```
rphillips, line 68: "v" and "w" could be more explicit... getFrameworkForValidatingWebhook and getFrameworkForMutatingWebhook. Thoughts?

pkg/webhook/pod_webhook.go
```diff
@@ -17,62 +17,219 @@ limitations under the License.
...
+	switch framework {
+	case "pod", "deployment", "statefulset":
+		return true
+	}
+	return false
+}
+
+func isClusterScopedFramework(framework string) bool {
+	switch framework {
+	case "kueue.x-k8s.io/clusterqueue", "kueue.x-k8s.io/resourceflavor", "kueue.x-k8s.io/cohort":
```
rphillips, line 102: might need constants for these at the top of the file?

pkg/webhook/pod_webhook.go
```diff
@@ -17,62 +17,219 @@ limitations under the License.
...
+//		      values: [true]
+//
+//		Merged selector:
+//		  matchLabels:
+//		    environment: prod
+//		  matchExpressions:
+//		    - key: kueue.openshift.io/managed
+//		      operator: In
+//		      values: [true]
+//		    - key: kubernetes.io/metadata.name
```
rphillips, line 144: I think this is a typo... metadata.namespace

rphillips:
It is worth adding to the README a blurb about the webhooks and the required namespace label.

pkg/webhook/pod_webhook.go
sohankunkerkar: The primary reason for keeping it this way is to avoid a few extra lines, but I'm not opposed to this idea

pkg/webhook/pod_webhook.go
sohankunkerkar: This line came from: https://github.com/openshift/kueue-operator/blob/main/bindata/assets/kueue-operator/validatingwebhook.yaml#L21

rphillips: Ok. metadata.name does correlate to the namespace.

rphillips:
lgtm. @kannon92 ?

PR #102, opened 2025-03-28
remove `make images` since we use operator-build

rphillips:
One place to create images. `make operator-build`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,7 +4,6 @@ all: build
 # Include the library makefile
 include $(addprefix ./vendor/github.com/openshift/build-machinery-go/make/, \
 	golang.mk \
-	targets/openshift/images.mk \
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -93,6 +84,10 @@ undeploy-ocp:
 deploy-cert-manager:
 	oc apply -f hack/manifests/cert-manager-rh.yaml
 
+.PHONY: deploy-upstream-cert-manager
```
kannon92, line 87: Why upstream and not Ocp cert?

kannon92 (approved):
/lgtm

one question but maybe upstream cert is used for local dev so i think its fine.

Makefile
rphillips: Right... It's for upstream work.

PR #107, opened 2025-04-01
remove assets as this is no longer needed

kannon92:
This is not used anymore.

Our assets are in bindata. I think this was remants of my original work but it was never removed.

cc @sohankunkerkar


sohankunkerkar (approved):
/lgtm

PR #112, opened 2025-04-02
Makefile: fetch channel and starting CSV based on OCP version

sohankunkerkar:


hack/deploy-cert-manager.sh
```diff
@@ -0,0 +1,78 @@
+#!/bin/bash
+set -eo pipefail
```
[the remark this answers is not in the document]
sohankunkerkar: Currently, the channel and starting CSV are hardcoded. It would be easier if users didn’t have to manually update them for every new OCP version, as they change with each release. This way, they could focus on development instead. For example, my `openshift-install` spins up a cluster with version 4.17.14, but if someone creates a new cluster using Cluster Bot with 4.18, they would need to adjust these values or at least keep track of them. This could be confusing, especially for someone new to the project.

hack/deploy-cert-manager.sh
```diff
@@ -0,0 +1,78 @@
...
+  labels:
+    name: cert-manager-operator
+---
+apiVersion: operators.coreos.com/v1
+kind: OperatorGroup
+metadata:
+  name: openshift-cert-manager-operator
+  namespace: cert-manager-operator
+spec:
+  targetNamespaces:
```
kannon92, line 63: https://docs.redhat.com/en/documentation/openshift_container_platform/4.18/html/security_and_compliance/cert-manager-operator-for-red-hat-openshift#cert-manager-install-cli_cert-manager-operator-install

This may get more challenging to support than we think.

sohankunkerkar: Tweaked this a bit to address the concern for pre 1.15 releases.

PR #115, opened 2025-04-02
WIP: detect openshift and enable local development

rphillips:


bindata/assets/kueue-operator/clusterroles/clusterrole-manager-role.yaml
```diff
@@ -17,6 +17,14 @@ rules:
   - patch
   - update
   - watch
+- apiGroups:
```
kannon92, line 20: Why would the operand need this?

We are creating the config map in the operator and I think we have those RBAC here.

sohankunkerkar:
@rphillips are you still working on this?

PR #122, opened 2025-04-03
[OCPNODE-3026] bundle-generate: preserve annotations/labels while generating a bundle

sohankunkerkar:


deploy/examples/rhoai_kueue.yaml
```diff
@@ -10,15 +10,8 @@ spec:
...
-      - "jobset.x-k8s.io/jobset"
-      - "kubeflow.org/mxjob"
-      - "kubeflow.org/paddlejob"
-      - "kubeflow.org/pytorchjob"
-      - "kubeflow.org/tfjob"
-      - "kubeflow.org/xgboostjob"
+      - BatchJob
+      - RayJob
+      - RayCluster
       externalFrameworks:
```
kannon92, line 16: Let's just remove this from RHOAI example.

This is no longer a valid YAML as appwrappers are moved into frameworks. But RHOAI does not support AppWrappers by default.

hack/preserve-bundle-labels.sh
```diff
@@ -0,0 +1,86 @@
+#!/bin/bash
+set -exou pipefail
+
+KUEUE_OPERAND_IMAGE="quay.io/redhat-user-workloads/kueue-operator-tenant/kubernetes-sigs-kueue@sha256:ec6688b483a2919d10d7ff6876f559c7b98961c6da425d80a653f0fe55db684d"
```
kannon92, line 4: Could we consider a different approach?

Maybe read from an environment variable?

This seems difficult to support programaticaly.

sohankunkerkar: I think `make bundle-generate` used to populate a `mustchange` value for all images. I believe we used to manually change them unless I'm missing something.

kannon92: My workload was to use `.envrc` and use something like `IMAGE_BUNDLE make bundle-generate`.

kannon92:
We can always iterate on this.

/lgtm
/approve

PR #161, opened 2025-04-08
Update our operator with the latest API changes

kannon92:


pkg/configmap/configmap.go
```diff
@@ -98,6 +98,35 @@ func buildLabelKeysCopy(labelKeys []kueue.LabelKeys) []string {
 	return ret
 }
 
+func buildManagedJobsWithoutQueueName(workloadManagement kueue.WorkloadManagement) bool {
+	if workloadManagement.LabelPolicy == "" {
```
rphillips, line 102: This if statement is superfluous, since the boolean comparison on 105 catches it.

pkg/configmap/configmap.go
```diff
@@ -98,6 +98,35 @@ func buildLabelKeysCopy(labelKeys []kueue.LabelKeys) []string {
...
+		return &configapi.WaitForPodsReady{Enable: false}
+	}
+}
+
+func buildFairSharing(preemption kueue.Preemption) *configapi.FairSharing {
+	switch preemption.PreemptionPolicy {
+	case kueue.PreemptionStrategyClassical:
+		return &configapi.FairSharing{Enable: false}
+	case kueue.PreemptionStrategyFairsharing:
+		return &configapi.FairSharing{Enable: true, PreemptionStrategies: []configapi.PreemptionStrategy{"LessThanOrEqualToFinalShare", "LessThanInitialShare"}}
```
rphillips, line 124: These strategies should be constants.

bundle/manifests/kueue-operator.clusterserviceversion.yaml
```diff
@@ -74,7 +74,6 @@ metadata:
     operatorframework.io/cluster-monitoring: "true"
     operatorframework.io/suggested-namespace: openshift-kueue-operator
     operators.openshift.io/valid-subscription: '["OpenShift Kubernetes Engine", "OpenShift Container Platform", "OpenShift Platform Plus"]'
-    operators.operatorframework.io/builder: operator-sdk-v1.33.0
```
sohankunkerkar, line 77: You might need to update this script as well: https://github.com/openshift/kueue-operator/blob/main/hack/preserve-bundle-labels.sh#L58

pkg/configmap/configmap.go
```diff
@@ -98,6 +98,35 @@ func buildLabelKeysCopy(labelKeys []kueue.LabelKeys) []string {
...
+	if workloadManagement.LabelPolicy == "" {
+		return false
+	}
+	return workloadManagement.LabelPolicy == kueue.LabelPolicyNone
+}
+
+func buildWaitForPodsReady(gangSchedulingPolicy kueue.GangScheduling) *configapi.WaitForPodsReady {
+	switch gangSchedulingPolicy.Policy {
+	case kueue.GangSchedulingPolicyNone:
+		return &configapi.WaitForPodsReady{Enable: false}
```
sohankunkerkar, line 111: How about this?
```go
func buildWaitForPodsReady(gangSchedulingPolicy kueue.GangScheduling) *configapi.WaitForPodsReady {
	if gangSchedulingPolicy.Policy == kueue.GangSchedulingPolicyByWorkload {
		return &configapi.WaitForPodsReady{Enable: true, BlockAdmission: ptr.To(false)}
	}
	return &configapi.WaitForPodsReady{Enable: false}
}
```

pkg/configmap/configmap.go
```diff
@@ -98,6 +98,35 @@ func buildLabelKeysCopy(labelKeys []kueue.LabelKeys) []string {
...
+	case kueue.GangSchedulingPolicyByWorkload:
+		return &configapi.WaitForPodsReady{Enable: true, BlockAdmission: ptr.To(false)}
+	default:
+		return &configapi.WaitForPodsReady{Enable: false}
+	}
+}
+
+func buildFairSharing(preemption kueue.Preemption) *configapi.FairSharing {
```
sohankunkerkar, line 119: Probably, we can follow the similar refactoring as `buildWaitForPodsReady` here

pkg/configmap/configmap.go
rphillips: I like the switch statement if more Policies are added.

kannon92: Yea I agree with @rphillips. We have BlockAdmission Parallel and Sequential actually so I may need to expand this now.

pkg/configmap/configmap.go
kannon92: updated.

pkg/configmap/configmap.go
kannon92: used kueue constants

sohankunkerkar:
/lgtm
@rphillips keeping it on hold for your review.
/hold

PR #162, opened 2025-04-09
pkg: use the configmap content for kueue deployment

sohankunkerkar:
This fix will help retain the settings we use from the ConfigMap and preserve the `internalCertManagement` value.

### DIFF
--- a/pkg/operator/target_config_reconciler.go
+++ b/pkg/operator/target_config_reconciler.go
@@ -357,7 +357,7 @@ func (c TargetConfigReconciler) sync() error {
 		return err
 	}
 
-	deployment, _, err := c.manageDeployment(kueue, specAnnotations, ownerReference)
+	deployment, _, err := c.manageDeployment(kueue, specAnnotations, ownerReference, cm)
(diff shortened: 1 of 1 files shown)


pkg/operator/target_config_reconciler.go
```diff
@@ -880,6 +880,36 @@ func (c *TargetConfigReconciler) manageDeployment(kueueoperator *kueuev1alpha1.K
...
+		Name: "kueue-config",
+		VolumeSource: v1.VolumeSource{
+			ConfigMap: &v1.ConfigMapVolumeSource{
+				LocalObjectReference: v1.LocalObjectReference{
+					Name: configMap.Name,
+				},
+			},
+		},
+	}
+	required.Spec.Template.Spec.Volumes = append(required.Spec.Template.Spec.Volumes, configVolume)
```
kannon92, line 894: I think that the existing deployment already has this no?

Should there be a replace rather than an append?

sohankunkerkar:
Not required.

PR #165, opened 2025-04-10
use CEL expression to limit fbc konflux jobs to just changes to fbc

cpmeadors:
This should only run bundle jobs on changes to bundle files, fbc jobs on changes to fbc files, and operator jobs on all other changes.  Testing is a bit tricky as there are many combinations of changes that need to work and this PR only covers one.  All jobs are running as expected since files for all three images changed.  The plan would be to merge and then check other PRs and make sure the right jobs are running.  If not we fix them or just revert this.

### DIFF
--- a/.tekton/kueue-operator-pull-request.yaml
+++ b/.tekton/kueue-operator-pull-request.yaml
@@ -9,6 +9,7 @@ metadata:
     pipelinesascode.tekton.dev/cancel-in-progress: "true"
     pipelinesascode.tekton.dev/max-keep-runs: "3"
     pipelinesascode.tekton.dev/on-cel-expression: event == "pull_request" && target_branch == "main"
+    pipelinesascode.tekton.dev/on-path-change-ignore: "[bundle/***, .tekton/kueue-bundle-pull-request.yaml, bundle.Dockerfile, fbc/***, .tekton/kueue-fbc-418-pull-request.yaml]"
   creationTimestamp:
   labels:
     appstudio.openshift.io/application: kueue-operator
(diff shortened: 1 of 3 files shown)


.tekton/kueue-operator-pull-request.yaml
```diff
@@ -9,6 +9,8 @@ metadata:
     pipelinesascode.tekton.dev/cancel-in-progress: "true"
     pipelinesascode.tekton.dev/max-keep-runs: "3"
     pipelinesascode.tekton.dev/on-cel-expression: event == "pull_request" && target_branch == "main"
+    pipelinesascode.tekton.dev/on-path-change-ignore: "[bundle/***, .tekton/kueue-bundle-pull-request.yaml, bundle.Dockerfile]"
+    pipelinesascode.tekton.dev/on-path-change-ignore: "[fbc/***, .tekton/kueue-fbc-418-pull-request.yaml]"
```
kannon92, line 13: You are defining the same key twice. I would be surprised that it actually appends this.

.tekton/kueue-operator-pull-request.yaml
```diff
@@ -9,6 +9,8 @@ metadata:
     pipelinesascode.tekton.dev/cancel-in-progress: "true"
     pipelinesascode.tekton.dev/max-keep-runs: "3"
     pipelinesascode.tekton.dev/on-cel-expression: event == "pull_request" && target_branch == "main"
+    pipelinesascode.tekton.dev/on-path-change-ignore: "[bundle/***, .tekton/kueue-bundle-pull-request.yaml, bundle.Dockerfile]"
```
kannon92, line 12:
```suggestion
    pipelinesascode.tekton.dev/on-path-change-ignore: "[bundle/***, .tekton/kueue-bundle-pull-request.yaml, bundle.Dockerfile, fbc/***, .tekton/kueue-fbc-418-pull-request.yaml]"
```

.tekton/kueue-operator-pull-request.yaml
cpmeadors: You are probably right.

kannon92 (commented):
/lgtm

PR #169, opened 2025-04-11
replace on-cel-expression with specific annotations

cpmeadors:
From Pipeline as Code docs:

"You still need to specify the event type and target branch. If you have a CEL expression the on-path-change-ignore annotation will be ignored."

So need to replace on-cel-expression with the simpler on-target-branch and on-event annotations to allow on-path-change and on-path-change-ignore to work.  Currently they do nothing according to the docs.


cpmeadors:
on-path-change-ignore overrides on-path-change.  This is a dead end.
/close

PR #193, opened 2025-04-14
README: Add project status badges

sohankunkerkar:


README.md
```diff
@@ -1,5 +1,10 @@
 # Kueue Operator
 
+[![Go Report Card](https://goreportcard.com/badge/github.com/openshift/kueue-operator)](https://goreportcard.com/report/github.com/openshift/kueue-operator)
+[![Quay Operator Image](https://img.shields.io/badge/operator-quay.io-blue?logo=docker)](https://quay.io/repository/redhat-user-workloads/kueue-operator-tenant/kueue-operator?tab=tags)
```
kannon92, line 4: Can we drop the quay images?

These are not the images we actually want to expose once we release.

sohankunkerkar: Yeah, that makes sense. I added a blurb at the beginning for those involved in the active development of the operator.

PR #214, opened 2025-04-16
Makefile: add e2e targets for prow CI

sohankunkerkar:
I ran these targets locally and both tests look fine.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -28,6 +28,10 @@ CODEGEN_OUTPUT_PACKAGE :=github.com/openshift/kueue-operator/pkg/generated
 CODEGEN_API_PACKAGE :=github.com/openshift/kueue-operator/pkg/apis
 CODEGEN_GROUPS_VERSION :=kueue:v1alpha1
 
+KUEUE_REPO := https://github.com/openshift/kubernetes-sigs-kueue.git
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -76,6 +80,9 @@ deploy-ocp:
 	oc apply -f deploy/crd/
 	oc apply -f deploy/examples/job.yaml
 	hack/revert-deploy-files.sh ${OPERATOR_IMAGE} ${KUEUE_IMAGE}
+	@echo "Waiting for Kueue Controller Manager..."
```
kannon92, line 83: What is meant by @echo or @timeout?

sohankunkerkar: https://www.gnu.org/software/make/manual/make.html#Echoing

kannon92:
LGTM.

What is your plan for integrating this with CI?

We merge this and then you update openshift/release to use these targets?

kannon92:
/hold

/lgtm

Hopefully you can answer https://github.com/openshift/kueue-operator/pull/214#issuecomment-2810093980 just so I understand.

sohankunkerkar:
> We merge this and then you update openshift/release to use these targets?

Yeah, that's correct.

kannon92:
tide seems stuck..

PR #222, opened 2025-04-17
Update to versioned  imageurls

cpmeadors:
This points the image urls for the operator and operand to the versioned images.


sohankunkerkar (approved):
/lgtm

PR #238, opened 2025-04-21
OCPNODE-3169: Kueue operand should use readOnlyRootFilesystem

kannon92:


kannon92:
/hold

there is some upstream problems with enabling this. gotta make sure its all good before merge.

PR #253, opened 2025-04-23
add network policy to isolate pod network traffic

tkashem:
- we assume that the operator namespace is shared with other components or operators, so the policy is limited to the operator pod and the kueue pod (operand) in the shared namespace
- if a policy applies to both the operator and the operand, then we use the label `app.openshift.io/name: kueue`. otherwise, if a policy applies to the operand only then we sue the label `app.kubernetes.io/name: kueue` 
- we introduce a deny-all policy that applies to both ingress and egress traffic for both the operator and the operand
- then we explicitly allow the expected ingress/egress traffic, this includes:
   - outgoing traffic to the kube-apiserver, 
   - egress to cluster DNS service, 
   - ingress traffic to metrics endpoint(s) published by either the operator, or the operand
   - ingress traffic to webhook (for the operand only)
   - ingress traffic to the visibility service (for the operand only)

### DIFF
--- a/bindata/assets/kueue-operator/controller-manager-metrics-service.yaml
+++ b/bindata/assets/kueue-operator/controller-manager-metrics-service.yaml
@@ -10,7 +10,7 @@ metadata:
   namespace: openshift-kueue-operator
 spec:
   ports:
-  - name: https
+  - name: metrics
     port: 8443
     protocol: TCP
     targetPort: 8443
--- a/pkg/operator/target_config_reconciler.go
+++ b/pkg/operator/target_config_reconciler.go
@@ -777,6 +777,17 @@ func (c *TargetConfigReconciler) manageNetworkPolicies(ownerReference metav1.Own
 		return fmt.Errorf("failed to read networkpolicy from directory %q: %w", networkPolicyDir, err)
 	}
 
+	// TODO: does the order of the creation of these policies matter?
+	// TODO: Since OLM does not support networkpolicy resource yet the
+	// operator Pod is creating policies for self isolation (in addition
+	// to operand isolation). let's say our operator creates the following
+	// network policy manifests for self and the operand in the following
+	// order: a) deny-all, b) allow-egress-api, c) allow egress cluster-dns,
+	// and d) allow-ingress-metrics; while creating these manifests in order,
+	// if there is a delay between a and b, long enough that deny-all takes
+	// effect and creation of b fails. If this can happen then the operator
+	// has lost access to the apiserver in a self inflicted manner. Should
+	// the operator create the deny-all policy last to avoid this issue?
 	for _, file := range files {
 		assetPath := filepath.Join(networkPolicyDir, file)
 		// TODO: move these resource helper functions to library-go
@@ -1150,7 +1161,7 @@ func (c *TargetConfigReconciler) manageServiceMonitor(ctx context.Context, kueue
 				{
 					Interval:        "30s",
 					Path:            "/metrics",
-					Port:            "https", // Name of the port you want to monitor
+					Port:            "metrics", // Name of the port you want to monitor
 					Scheme:          "https",
 					BearerTokenFile: "/var/run/secrets/kubernetes.io/serviceaccount/token",
 					TLSConfig: &monitoringv1.TLSConfig{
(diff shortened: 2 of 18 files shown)


pkg/operator/target_config_reconciler.go
```diff
@@ -311,6 +311,11 @@ func (c TargetConfigReconciler) sync() error {
 		return err
 	}
 
+	if err := c.manageNetworkPolicies(ownerReference); err != nil {
```
kannon92, line 314: We would need to change the namespace if a user installs this in a different namespace from `openshift-kueue-operator`.

pkg/operator/networking_test.go
```diff
@@ -0,0 +1,234 @@
...
+			}
+			if want, got := test.err, err; want != got {
+				t.Errorf("expected err to be: %v, but got: %v", want, got)
+			}
+			if want, got := test.operations, operationsObserved; !cmp.Equal(want, got) {
+				t.Errorf("expected operations to match, diff: %s", cmp.Diff(want, got))
+			}
+
+			// the original spec in the desired object passed to ApplyNetworkPolicy should
+			// be immutable, mutating the object returnd from ApplyNetworkPolicy
```
kannon92, line 204:
```suggestion
			// be immutable, mutating the object returned from ApplyNetworkPolicy
```

bindata/assets/kueue-operator/networkpolicy/allow-ingress-visi
```diff
@@ -0,0 +1,15 @@
+apiVersion: networking.k8s.io/v1
```
cpmeadors, line 1: Did the filename of this get truncated?  Looks like it wanted to be allow-ingress-visibility.yaml based on how the other files are named.

tkashem: yes, not sure how it happened, let me fix it, good catch :)

test/e2e/e2e_operator_test.go
```diff
@@ -65,6 +66,28 @@ const (
...
+		}
+	}
+	return pods
+}
+
+func findKueuePods(list *corev1.PodList) []*corev1.Pod {
+	pods := make([]*corev1.Pod, 0)
+	for i := range list.Items {
+		pod := &list.Items[i]
+		if strings.HasPrefix(pod.Name, "kueue-controller-manager") {
```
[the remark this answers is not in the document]
tkashem: I am not sure, I think the name of the kueue pod starts with `kueue-controller-manager` - https://github.com/tkashem/kueue-operator/blob/5565cbb70805760618317805366c742da0349b85/test/e2e/e2e_operator_test.go#L92-L101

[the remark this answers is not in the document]
haircommander: hmm I wonder what else is found that has kueue prefix? I would expect a network policy to apply to all given how you filter for the app name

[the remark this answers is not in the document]
tkashem: we are operating with the constraint that the operator may share a namespace with other operator(s). otherwise, we could leave the `podSelector` empty in which case all pods in the non-shared namespace would be selected.
with the shared namespace approach, for any future kueue pods, we need to add the label so it is matched.

I am not sure why an admin would choose two separate operators to share a namespace, that seems to carry some security concerns, they may also infringe on one another. we can have an offline discussion on this topic.

bindata/assets/kueue-operator/controller-manager-metrics-service.yaml
```diff
@@ -10,7 +10,7 @@ metadata:
   namespace: openshift-kueue-operator
 spec:
   ports:
-  - name: https
+  - name: metrics
```
kannon92, line 13: can you make sure to do a manual test that metrics still work with OCP?

Deploy operator, make sure that the metrics dashboard still shows Kueue metrics.

I find that these names can be subtle so just want to make sure this actually works.

PannagaRao: I verified this yesterday and the metrics endpoint is available in the metrics dashboard.

kannon92:
/approve

@haircommander @cpmeadors @PannagaRao any outstanding items on this?

If not, please feel free to add your lgtm.

kannon92:
Can you retarget this to release-1.0?

cpmeadors:
Main is defunct until we are done with the 1.0 release.  Holding until retargeted to release-1.0
/hold

tkashem:
/close

(https://github.com/openshift/kueue-operator/pull/342)

PR #258, opened 2025-04-24
add operator image back to csv

cpmeadors:
Bundle generation process was adding the wrong image urls for the operator and operand.  These must be the prod repos (registry.redhat.io) and use digests.  This is required because the bundle is not rebuilt on release.  A consequence of this is that you need to use a mirror set to test with the bundle.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -78,14 +78,11 @@ get-kueue-image:
 	@rm -f .kueue_commit_id
 
 .PHONY: bundle-generate
-bundle-generate: operator-sdk regen-crd manifests get-kueue-image
-	@KUEUE_IMAGE=$$(cat .kueue_image); \
+bundle-generate: operator-sdk regen-crd manifests
 	hack/update-deploy-files.sh ${OPERATOR_IMAGE} $$KUEUE_IMAGE
 	${OPERATOR_SDK} generate bundle --input-dir deploy/ --version ${OPERATOR_VERSION}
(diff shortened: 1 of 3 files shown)


kannon92:
should this go to main? and then cherry-picked?

Makefile
```diff
@@ -78,14 +78,11 @@ get-kueue-image:
 	@rm -f .kueue_commit_id
 
 .PHONY: bundle-generate
-bundle-generate: operator-sdk regen-crd manifests get-kueue-image
-	@KUEUE_IMAGE=$$(cat .kueue_image); \
```
sohankunkerkar, line 82: Can you add some description to the GitHub PR? It looks like we are no longer using a kueue-0-11 image for bundle?

cpmeadors: I updated the PR description.  All the kueue images will get pushed to the same repo.  The digest will point to a build from kueue-0-11.

cpmeadors:
> should this go to main? and then cherry-picked?

Probably, but I just tried to change the base branch to main and ran into a mess of conflict.

cpmeadors:
> > should this go to main? and then cherry-picked?
> 
> Probably, but I just tried to change the base branch to main and ran into a mess of conflict.

I recreated these changes on main. https://github.com/openshift/kueue-operator/pull/259  we can use that and cherry pick back to release-1.0 or whatever you think it best.

kannon92:
can you rebase this?

PR #259, opened 2025-04-25
fix bundle generation and csv

cpmeadors:


sohankunkerkar:
Is it possible to add `make bundle-generate` to prow CI? It will help us verifying next time if we try generating something,

kannon92:
> Is it possible to add `make bundle-generate` to prow CI? It will help us verifying next time if we try generating something,

Yea, we have a generate task already. We can add something there.

PR #266, opened 2025-04-28
use 0.1.0 channel for tech preview

kannon92:
Update our channels to reflect 0.1.0 release.


cpmeadors:
I just realized that these files should be generated when the sha is updated in the template.  These values are extracted from the actual bundle reference there.  We should not be modifying the fbc fragments directly.  Sames as we should not be modifying the CSV directly since it is generated as well.

PR #280, opened 2025-04-30
reorder CSV to avoid conflicts

cpmeadors:
Replaces PR #276 and resolve the conflict for future kueue image nudges

Resolves: [OCPNODE-3208](https://issues.redhat.com//browse/OCPNODE-3208)


bundle/manifests/kueue-operator.clusterserviceversion.yaml
```diff
@@ -298,11 +298,11 @@ spec:
                   valueFrom:
                     fieldRef:
                       fieldPath: metadata.annotations['olm.targetNamespaces']
+                - name: RELATED_IMAGE_OPERAND_IMAGE
```
kannon92, line 301: Could you do this change without changing the sha at least?

kannon92: /hold

PR #296, opened 2025-05-02
change cel expression to eliminate redundant builds

cpmeadors:
Already updated main, but this will eliminate the redundant builds that are causing circular nudges.


cpmeadors:
/override "Red Hat Konflux / kueue-operator-1-0-enterprise-contract / kueue-bundle-1-0"
/override "Red Hat Konflux / kueue-operator-1-0-enterprise-contract / kueue-operator-1-0"
/override "Red Hat Konflux / kueue-operator-1-0-enterprise-contract / pr group"

PR #298, opened 2025-05-02
tweak cel expression for FBCs

cpmeadors:
This is temporary until we move FBCs to their own repo, but still need to limit extra operator builds to eliminate extra nudges.


cpmeadors:
/override "Red Hat Konflux / kueue-operator-enterprise-contract / kueue-operator"

PR #312, opened 2025-05-07
Fix ec for must gather

cpmeadors:
enable hermetic and source-image-build
Add required labels to containerfile


kannon92:
I was afraid of this.

> Base image "quay.io/openshift/origin-must-gather" is from a disallowed registry

PR #317, opened 2025-05-08
use latest for ubi-micro

cpmeadors:
trying latest tag to see if this resolves source image signature error


bundle.Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM registry.access.redhat.com/ubi9/ubi-micro@sha256:84bb350a205a892350d995177f51c81180397d63f1028634623a0c487837f90c
+FROM scratch
```
kannon92, line 1: scratch will fail fips check.

micro is the correct choice here.

cpmeadors: works fine on other operators.  Why would it fail?

kannon92: Hmm. Maybe you are correct. I do skip the ecosystem check for bundles now.

It was failing the ecosystem check before.

PR #342, opened 2025-05-21
add network policy to isolate pod network traffic

tkashem:
- we assume that the operator namespace is shared with other components or operators, so the policy is limited to the operator pod and the kueue pod (operand) in the shared namespace
- if a policy applies to both the operator and the operand, then we use the label `app.openshift.io/name: kueue`. otherwise, if a policy applies to the operand only then we sue the label `app.kubernetes.io/name: kueue` 
- we introduce a deny-all policy that applies to both ingress and egress traffic for both the operator and the operand
- then we explicitly allow the expected ingress/egress traffic, this includes:
   - outgoing traffic to the kube-apiserver, 
   - egress to cluster DNS service, 
   - ingress traffic to metrics endpoint(s) published by either the operator, or the operand
   - ingress traffic to webhook (for the operand only)
   - ingress traffic to the visibility service (for the operand only)

### DIFF
--- a/bundle/manifests/kueue-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/kueue-operator.clusterserviceversion.yaml
@@ -287,6 +287,7 @@ spec:
           template:
             metadata:
               labels:
+                app.openshift.io/name: kueue
                 name: openshift-kueue-operator
             spec:
               containers:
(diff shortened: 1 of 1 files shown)


deploy/02_role.yaml
```diff
@@ -0,0 +1,18 @@
+kind: Role
```
kannon92, line 1: If you are changing these values you need to regenerate the bundle.

kannon92: https://github.com/openshift/kueue-operator?tab=readme-ov-file#operator-bundle-development

PR #346, opened 2025-05-29
OSDOCS-14688: Update Operator description and other details

abrennan89:


bundle/manifests/kueue-operator.clusterserviceversion.yaml
```diff
@@ -125,8 +125,9 @@ spec:
     - kind: Kueue
       name: kueues.operator.openshift.io
       version: v1alpha1
-  description: Kueue Operator description. TODO.
-  displayName: Kueue Operator
+      description: |
+        Create a `Kueue` custom resource (CR) to configure your installation.
```
abrennan89, line 129: This should probably be "CRD" (CustomResourceDefinition) - but I wasn't sure. We had the conversation previously about how it's maybe a weird UX that users need to create this one instead of just the other CRs for queues etc. Idk if we plan to change this UX.

kannon92:
Why was this closed?

PR #354, opened 2025-06-04
OCPNODE-3358: add description to RHBok

kannon92:


hack/preserve-bundle-labels.sh
```diff
@@ -7,6 +7,7 @@ DESIRED_BASE="scratch"
 CSV_FILE="bundle/manifests/kueue-operator.clusterserviceversion.yaml"
 DOCKERFILE="bundle.Dockerfile"
 DISPLAY_NAME="Red Hat Build of Kueue"
```
abrennan89, line 9:
```suggestion
DISPLAY_NAME="Red Hat build of Kueue"
```

hack/preserve-bundle-labels.sh
```diff
@@ -65,6 +66,9 @@ sed -i "s|image: mustchange|image: ${KUEUE_OPERATOR_IMAGE}|g" "${CSV_FILE}"
 # Replace Display Name
 sed -i "s|displayName: Kueue Operator|displayName: ${DISPLAY_NAME}|" "${CSV_FILE}"
 
+# Replace Description
+sed -i "s|description: Kueue Operator description. TODO.|description: ${DESCRIPTION}|" "${CSV_FILE}"
```
abrennan89, line 70: Is there a way to add a description for the Kueue CR as well?

kannon92: I will do this as a follow up.

I think I can use olm.

sohankunkerkar (approved):
/lgtm

kannon92:
@everettraven is helping me with looking into the generate failures.

I can reproduce this outside of this PR so I think something else is going on.

kannon92:
/override ci/prow/test-generate

We have #356 up which will fix this CI failure. Going to override.

PR #357, opened 2025-06-06
OCPNODE-3306: Add `ManagedJobsNamespaceSelector` to Kueue configuration

MaysaMacedo:
When using `labelPolicy` with `None` value, the `managedJobsWithoutQueueName` is enabled, consequently any job that doesn't have the kueue label set will still be managed by Kueue. To reduce the scope of affected jobs this commits set the `ManagedJobsNamespaceSelector` to only select the namespaces with label `kueue.openshift.io/managed=true`.


kannon92:
LGTM

I'll wait for monday to give final approvals.

/ok-to-test

PR #371, opened 2025-06-16
update bundle to latest

kannon92:


kannon92:
We will do a 0.2.0 release of operator before we update the FBC.

PR #375, opened 2025-06-18
Remove resource limits from kueue-controller-manager

MaysaMacedo:


MaysaMacedo:
/hold

I want see CI results then will update commit message.

pkg/operator/target_config_reconciler.go
```diff
@@ -953,6 +954,14 @@ func (c *TargetConfigReconciler) manageDeployment(kueueoperator *kueuev1.Kueue,
 		},
 	}
 
+	required.Spec.Template.Spec.PriorityClassName = "system-cluster-critical"
+	required.Spec.Template.Spec.Containers[0].Resources = v1.ResourceRequirements{
+		Requests: v1.ResourceList{
+			v1.ResourceCPU:    resource.MustParse("500m"),
```
MaysaMacedo, line 960: @kannon92 I'm enforcing the same values from here https://github.com/openshift/kueue-operator/blob/main/bindata/assets/kueue-operator/deployment.yaml#L59-L61

Are those the expected values?

kannon92: I think that is fine.

kannon92 (changes requested):
Change looks good but you will need to generate a bundle with your changes.

you can generate a bundle with make bundle-generate 
But you will need to set the environment variables to the same images in the bundle otherwise you overwrite the nudge images

I’ve been meaning to get around to add a way to catch issues where bundle changes are needed but they weren’t. Maybe a follow up?

MaysaMacedo:
@kannon92 I have issues with defining which environment variables should I set. Do you know which ones is it?
and should the values match what is defined [here](https://github.com/openshift/kueue-operator/blob/main/bundle/manifests/kueue-operator.clusterserviceversion.yaml#L304-L305)?

Also, should we always use operator-sdk version as [v1.37.0](https://github.com/openshift/kueue-operator/blob/main/bundle.Dockerfile#L9)? If yes, I will set a default value to [OPERATOR_SDK_VERSION](https://github.com/openshift/kueue-operator/blob/main/Makefile#L167C100-L167C120).

kannon92:
> @kannon92 I have issues with defining which environment variables should I set. Do you know which ones is it? and should the values match what is defined [here](https://github.com/openshift/kueue-operator/blob/main/bundle/manifests/kueue-operator.clusterserviceversion.yaml#L304-L305)?
> 
> Also, should we always use operator-sdk version as [v1.37.0](https://github.com/openshift/kueue-operator/blob/main/bundle.Dockerfile#L9)? If yes, I will set a default value to [OPERATOR_SDK_VERSION](https://github.com/openshift/kueue-operator/blob/main/Makefile#L167C100-L167C120).

You should match the images in release-1.0 bundle/manifests/kueue-operator.clusterserviceversion.yaml#L304-L305).

If you want to find which ones to use, you can go to the catalog and get the latest sha.

Just use those images for OPERATOR_IMAGE and KUEUE_IMAGE.

kannon92:
It may be worth having a e2e test that verifies this. Maybe check deployment for priorityClass and resources.

kannon92:
/lgtm
/approve

https://github.com/openshift/kueue-operator/pull/375#issuecomment-2988979618 can be a follow up though.

PR #382, opened 2025-06-19
remove fbcs from repo. using openshift/kueue-fbc for now on

kannon92:


cpmeadors:
There is some content in the README that need to be removed as well.

PR #383, opened 2025-06-19
remove fbc from release-1.0. use openshift/kueue-fbc for now on

kannon92:


cpmeadors:
/hold
There is some content in the README about the FBC that should be removed as well.

PR #407, opened 2025-06-26
test/e2e: address e2e test failure for statefulset

sohankunkerkar:
This change prevents conflicts when both Deployments and StatefulSets exist in the same namespace and ensures tests can reliably identify the correct Pods. It also fixes the issue where the Stateful workload lacked the kueue.x-k8s.io/job-uid label that the test was looking for—instead, it now uses an ownerReferences field that points to the Pod.


sohankunkerkar:
cc @PannagaRao

PannagaRao:
I see that the ci/prow/test-e2e-4-18 is still failing due to the managed statefulset test case. Are we expecting this PR to fix that ?

PR #415, opened 2025-06-30
remove deploy-ocp from upstream tests

kannon92:


sohankunkerkar:
I think you can remove it from the upstream tests make target, but it would still be useful to have it as a standalone make target.

MaysaMacedo (approved):

PannagaRao:
I think we will need to explicitly create the Kueue CR when we are removing the deploy-ocp target.

PR #420, opened 2025-06-30
hack: wait for the packagemanifests to become ready

sohankunkerkar:


sohankunkerkar:
I think only one failure.. which means the flake is still there.

sohankunkerkar:
> > I think only one failure.. which means the flake is still there.
> 
> I see. Maybe we bump the timeout to 10 minutes and then we merge this?

I added a retry mechanism. There's a race condition or timing issue where the package manifest becomes unavailable right after the loop detects it.

kannon92 (commented):
/hold cancel

PR #428, opened 2025-07-02
disable coverity for now

kannon92:


kannon92:
No bundle pipeline so these tests will fail.

/override ci/prow/test-e2e-4-17
/override ci/prow/test-e2e-upstream-4-17

PR #437, opened 2025-07-07
test/e2e: don't install cr if already exists

sohankunkerkar:
This is required for running e2e tests on the disconnected clusters where we already apply the CR.


sohankunkerkar:
not required..  I think this should be handled in openshift/release itself.

PR #439, opened 2025-07-07
NO-JIRA: Clean up configmap & secrets on deleting Kueue CR

PannagaRao:

### DIFF
--- a/pkg/operator/target_config_reconciler.go
+++ b/pkg/operator/target_config_reconciler.go
@@ -186,6 +186,7 @@ func (c TargetConfigReconciler) sync() error {
 			c.cleanUpCertificatesAndIssuers,
 			c.cleanUpClusterRoles,
 			c.cleanUpClusterRoleBindings,
+			c.cleanUpResources,
 		}
 
 		for _, step := range cleanupResources {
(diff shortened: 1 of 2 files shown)


pkg/operator/target_config_reconciler.go
```diff
@@ -449,6 +450,54 @@ func (c *TargetConfigReconciler) removeFinalizerFromKueueInstance(kueue *kueuev1
 	return c.updateFinalizer(kueue, false)
 }
 
+func (c *TargetConfigReconciler) cleanUpResources(ctx context.Context) error {
```
kannon92, line 453: one suggestion: 

it may be worth separating these into separate functions.

cleanupResources is a general function and we are deleting different resources here.

kannon92 (commented):
one nit.

You can do this as a followup.

/lgtm
/approve

kannon92:
/jira refresh

I'm not sure why OCPBugs doesn't like this but we don't really need it.

pkg/operator/target_config_reconciler.go
PannagaRao: Yeah we could do that. Since there were just 3 resources I grouped them but can break them up.

PR #440, opened 2025-07-08
OCPNODE-3449: Remove assets generated during the e2e test

MaysaMacedo:
Even though we already have a destroy of namespace
where k8s assets are created, there is not wait for
all the assets to be properly cleaned up. After triggering
deletion of namespace we already proceed to the next
test scenario which is of deleting Kueue instance.
This can leave some CRDs behind if some k8s assets
were not yet fully cleaned up, and even when we recreate
the Kueue instance in the following test scenario,
this CRD will eventually get deleted.
This commit adds defer clean ups to assets generated
in the e2e test to ensure no leftovers are left by the
OCP e2e suite that could affect other suites that run after it.


kannon92:
/lgtm
/approve

Looks like it got us past the issue. The failures seem unrelated to not finding the resources.

PR #448, opened 2025-07-10
Fix clusterRole clean up

MaysaMacedo:
We shouldn't be deleting cluster roles
that are meant for the operator and not
the operand, even if they get recreated.
This commit fixes the issue by skipping
the clean up of any cluster role that
has openshift.io in the name.


kannon92 (approved):
/lgtm

PR #493, opened 2025-07-21
Update bundle pipeline to be single arch

cpmeadors:
Copy of #491 to update pipeline and set hermetic and source to true to pass EC


kannon92:
Going to force merge this.

ec are still a work in progress.

PR #509, opened 2025-08-01
WIP: refactor the operator to manage kueue via git submodules thoughts

rphillips:
I've been pondering this, and believe this approach would serve us well. Instead of managing kueue out of the openshift/kubernetes-sigs-kueue repo, we could manage it via a git submodule. There are a few benefits to this:

1. Automation becomes much much easier
    - Konflux will autonudge git submodules, so when there is a commit to the 'main' branch upstream, or 'release-0.11', etc... the commit will be reflected with a nudge in konflux within this repo. CI can run tests on that PR, etc. The CI can auto-merge that bump automatically when the tests pass.
2. Takes a bunch of maintenance out of maintaining some sort of bot to do the syncing in kubernetes-sigs-kueue. Konflux/Mintmaker does it for us.

When a new release of kueue is branched, we create the git submodule within the operator. CI runs.

We would have to refactor some of the dependency management we have, but would be worth it. The opreator can test upgrades between different kueue releases as well.

The operator would need to be taught a few things. Deploy 'main', 'release-0.11' or 'release-0.12' release of kueue, or upgrade to this release of kueue (we likely already have to do this anyway in some fashion). To clarify, the operator will likely not need a user to specify this version (so no API change). If we find through testing that 0.11 and 0.12 are not breaking, then the opreator upgrades them to 0.12. We can figure this out later though since the opreator itself could be semver'ed with what kueue version it supports (1.0 operator == release-0.11, 1.2 operator == release-0.12, etc).

The maintenance would be simplified greatly and we would have continuous integration with changes upstream for basically free (unless upstream breaks something).

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,11 @@
+[submodule "upstream/main/src"]
+	path = upstream/main/src
+	url = https://github.com/kubernetes-sigs/kueue.git
+[submodule "upstream/release-0.11/src"]
+	path = upstream/release-0.11/src
+	url = https://github.com/kubernetes-sigs/kueue.git
+	branch = release-0.11
+[submodule "upstream/release-0.12/src"]
+	path = upstream/release-0.12/src
+	url = https://github.com/kubernetes-sigs/kueue.git
+	branch = release-0.12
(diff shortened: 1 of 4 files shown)


kannon92:
cc @sohankunkerkar @MaysaMacedo @PannagaRao

kannon92:
> Yes. We would build the images in konflux.

How does that work with the submodule where you are just using the upstream changes?

Wouldn't we want a fork building these for each release?

PR #515, opened 2025-08-05
OCPNODE-3537: use main kueue for operator e2e

kannon92:
Update release version to 1.1 and use main images for e2e.


kannon92:
#531 

It seems we don't use the bundle on the PR but the bundle that is last published. This is causing some problems because I think the old image is a release-0.11 image while I am trying to get the image to be release-0.12.

To "fix" this for now, I decided to move bundle changes to #531 and we should merge that first if CI passes.

PR #528, opened 2025-08-14
pkg/configmap: add upstream resource transformations support

sohankunkerkar:
Add Resources field to KueueConfiguration to expose upstream Kueue's resource transformation feature. This allows users to configure resource exclusions and transformations (Replace/Retain strategies) via the operator CR.


kannon92 (commented):
/hold

The main reason why we went through API review was to not put a dependency on the kueue API in our API.

It would be worth going through an api-review for this new API.

cc @JoelSpeed @everettraven

pkg/apis/kueueoperator/v1/types.go
```diff
@@ -70,6 +71,11 @@ type KueueConfiguration struct {
 	// This default could change over time.
 	// +optional
 	Preemption Preemption `json:"preemption"`
+	// resources controls how Kueue manages resource transformations.
+	// Resource transformations allow converting PodSpec resources into Workload resource requests.
+	// This field is optional.
+	// +optional
+	Resources *configapi.Resources `json:"resources,omitempty"`
```
JoelSpeed, line 78: We recommend not to import types from other APIs. You cannot control the features added to the API and implicitly support anything that the upstream project adds without any form of internal review or testing.

I would suggest bringing the types into the project here directly from the upstream and then we can review them to make sure they conform to our API standards, and that we are happy supporting them in the long term.

Bear in mind that the upstream types are v1beta1 and so could also be subject to change over time, where our APIs are v1 and cannot change in the same ways

sohankunkerkar: Yeah, this makes sense to me. I have created an upstream issue here: https://github.com/kubernetes-sigs/kueue/issues/6582 to understand the future of that feature.

sohankunkerkar: ok, Should I propose the change in the openshift enhancement first?

kannon92: So the process I went through was to propose the full API in openshift/api.

The linter helps get most of the API standards and then we do the review there.

And then we copy those api changes back into this repo.

This is a bit painful and I'd prefer to maybe just have the API review done on the operator repo if possible.

kannon92: I avoided having the API in the enhacement as it was being discussed in the API review anyway.

kannon92:
Can you add this to main actually?

pkg/apis/kueueoperator/v1/types.go
JoelSpeed: It is probably worth setting up the API linting for this repo as well

PR #557, opened 2025-08-26
controller sync cleanup

rphillips:
* queries the cluster if ServiceMonitors are supported
* uses the library-go controller factory to register informers
* fixes the sync loop being called too often
* registers more specAnnotations to prevent too many updates


pkg/operator/target_config_reconciler.go
```diff
@@ -74,9 +74,11 @@ type TargetConfigReconciler struct {
 	queue                      workqueue.TypedRateLimitingInterface[queueItem]
 	kubeInformersForNamespaces v1helpers.KubeInformersForNamespaces
 	crdClient                  apiextv1.ApiextensionsV1Interface
+	crdInformer                apiextinformer.SharedInformerFactory
 	operatorNamespace          string
 	resourceCache              resourceapply.ResourceCache
 	kueueImage                 string
+	serviceMonitorSupport      bool
```
kannon92, line 81: I don't really understand the motivation between serviceMonitorSupport.

Is this to deploy on a non ocp cluster?

pkg/operator/target_config_reconciler.go
```diff
@@ -137,11 +142,42 @@ func NewTargetConfigReconciler(
...
+		kubeInformersForNamespaces.InformersFor(c.operatorNamespace).Rbac().V1().ClusterRoles().Informer(),
+		kubeInformersForNamespaces.InformersFor(c.operatorNamespace).Rbac().V1().RoleBindings().Informer(),
+		kubeInformersForNamespaces.InformersFor(c.operatorNamespace).Rbac().V1().Roles().Informer(),
+		kubeInformersForNamespaces.InformersFor(c.operatorNamespace).Networking().V1().NetworkPolicies().Informer(),
+	).ResyncEvery(5*time.Minute).
+		WithSync(c.sync).
+		ToController("KueueOperator", c.eventRecorder), nil
+}
 
+func (c TargetConfigReconciler) sync(ctx context.Context, syncCtx factory.SyncContext) error {
```
kannon92, line 180: Neither of these function parameters are used.

PR #607, opened 2025-09-04
fix vanguard konflux failure

kannon92:


kannon92:
going to force merge to fix these EC failures.

PR #762, opened 2025-09-29
add konflux integration test

rphillips:


.tekton/integration/integration-test.yaml
```diff
@@ -0,0 +1,192 @@
+kind: Pipeline
+apiVersion: tekton.dev/v1beta1
+metadata:
+  name: integration-tests
```
kannon92, line 4: Do we control what version of OCP this runs on?

rphillips: 4.19

kannon92:
The code looks good to me. But the test is still failing.

.tekton/integration/integration-test.yaml
```diff
@@ -0,0 +1,199 @@
...
+            echo "Images to be tested:"
+            echo "Kueue operator: ${OPERATOR_IMAGE}"
+            echo "Kueue operand:  ${OPERAND_IMAGE}"
+            echo "Kueue must-gather:  ${MUSTGATHER_IMAGE}"
+            echo "----------------------------"
+
+            git clone $SRC_REPO
+            pushd $SRC_BASEDIR
+            git checkout $SRC_REVISION
+            make deploy-upstream-cert-manager
```
kannon92, line 157: Why not use OCP cert manager?

.tekton/integration/integration-test.yaml
```diff
@@ -0,0 +1,199 @@
...
+
+            oc create ns ${NAMESPACE}
+            oc apply -f deploy/
+            oc apply -f deploy/crd
+            popd
+
+            # wait for operator
+            timeout 300s bash -c 'until oc get deployment openshift-kueue-operator -n openshift-kueue-operator -o jsonpath="{.status.conditions[?(@.type==\"Available\")].status}" | grep -q "True"; do sleep 10; echo "Still waiting..."; done'
+
+            if make e2e-ci-test; then
```
kannon92, line 177: This only runs our operator e2e.

We also have https://github.com/openshift/kueue-operator/blob/main/Makefile#L214.

PR #803, opened 2025-10-07
Update Kueue install mode to OwnNamespace

MaysaMacedo:
Right now the CSV permissions are "promoted" to ClusterRoles when the operator is installed with AllNamespaces mode, so let's avoid having a clusterRole and role being created for the same resource, as an example network policy.
Also, our operator does not need to watch events on all namespaces, so let's change it to OwnNamespace.


kannon92 (commented):
/lgtm
/approve

PR #822, opened 2025-10-14
Fix integration tests

rphillips:
I fixed the Kueue Operator integration tests in this pr. There are few things to note:

* I added labels to the suites, so that we can easily skip them if necessary.
* The Ginkgo suites run in parallel, by default. Since the suites are not idempotent (using random naming for things), they stomp on each other causing 'resource exists' or 'deletion' errors.
* It("") test suits also need to cleanup up after themselves and be idemptotent.

For now, I have disabled the qe tests since they are using hard coded names from yaml files. Additionally, I have skipped the kueue-operator teardown ("disruptive") tests. We will need to followup and add some user stories around fixing these things.


rphillips:
/hold cancel

Let's get this in if it passes.

rphillips:
/override "Red Hat Konflux / integration-test / pr group"

PR #836, opened 2025-10-17
OCPKUEUE-336: Add support to kueue.openshift.io/allow-nominal-concurrency-shares-update annotation

MaysaMacedo:
This commit adds support to an annotation to the `PriorityLevelConfiguration` asset, which allows the user to modify the `nominalConcurrencyShares` field to values bigger or equal than 2 and less or equal than 5 and to 0, which rejects any traffic to the extension API. This is needed in case the user experiences issues with the API and want to tune it. Also, note that the default value is being modifed to 2, which means this would allow 32 requests to be handled concurrently.

### DIFF
--- a/test/e2e/testutils/utils.go
+++ b/test/e2e/testutils/utils.go
@@ -413,10 +413,6 @@ func CleanUpWorkload(ctx context.Context, kueueClient *upstreamkueueclient.Clien
 // CleanUpKueueInstance deletes the specified Kueue instance and waits for its removal.
 func CleanUpKueueInstance(ctx context.Context, kueueClientset *kueueclient.Clientset, name string) {
 	By(fmt.Sprintf("Destroying Kueue %s", name))
-	removeFinalizersWithPatch(func() error {
-		_, err := kueueClientset.KueueV1().Kueues().Patch(ctx, name, types.MergePatchType, removeFinalizersMergePatch, metav1.PatchOptions{})
-		return err
-	})
 	err := kueueClientset.KueueV1().Kueues().Delete(ctx, name, metav1.DeleteOptions{})
 	Expect(err).NotTo(HaveOccurred())
 	Eventually(func() error {


pkg/util/resourceapply/flowcontrol.go
```diff
@@ -109,6 +109,13 @@ func ApplyPriorityLevelConfiguration(ctx context.Context, getter flowcontrolclie
 		return current, false, nil
 	}
 
+	if current.Annotations != nil && current.Annotations["adapt-nominal-concurrency-shares"] == "true" {
```
kannon92: Usually annotations have a prefix that corresponds to the org defining it.

Is this annotation something we are adding as part of our operator?

Or is this an existing annotation?

MaysaMacedo: yes, I'm working on the e2e and this will also be documented.

kannon92: maybe we can follow our API naming for Kueue. 

something like `kueues.kueue.openshift.io/adapt-nominal-concurrency-shares`.

That way its clearly coming from our API and not an external annotation.

test/e2e/e2e_visibility_on_demand_test.go
```diff
@@ -0,0 +1,150 @@
...
+		testQueue := "test-queue"
+
+		var cleanupClusterQueue, cleanupLocalQueue, cleanupResourceFlavor func()
+		var err error
+		var ns *corev1.Namespace
+		var priorityClient flowcontrolclientv1.PriorityLevelConfigurationInterface
+
+		BeforeAll(func(ctx context.Context) {
+			priorityClient = clients.KubeClient.FlowcontrolV1().PriorityLevelConfigurations()
+			ns, err = kubeClient.CoreV1().Namespaces().Create(context.TODO(), &corev1.Namespace{
```
kannon92: Should we have a create and wait for object to exist?

I don't think create is a blocking call.

test/e2e/e2e_visibility_on_demand_test.go
```diff
@@ -0,0 +1,155 @@
...
+			Expect(err).NotTo(HaveOccurred())
+
+			By("Check is able to access the pending workload")
+			_, err = visibilityClient.LocalQueues(ns.Name).GetPendingWorkloadsSummary(ctx, testQueue, metav1.GetOptions{})
+			Expect(err).NotTo(HaveOccurred())
+		})
+		It("should not allow modification of the nominal concurrency shares to 1", func(ctx context.Context) {
+			By("Modifying the PriorityLevelConfiguration with nominal concurrency shares set to 1")
+			updateNominalConcurrencyShares(ctx, priorityClient, 1)
+			By("Wait to verify the value of nominal concurrency shares is changed back to the default")
```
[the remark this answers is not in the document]
MaysaMacedo: given default value is 2, we allow the range of 2-5 and 0 to disable it.

PR #850, opened 2025-10-22
Define specAnnotation about visibilityOnDemand assets

MaysaMacedo:
Define the specAnnotation for the vibilityOnDemand assets to follow the same pattern as the other assets that are managed by Kueue operator.

### DIFF
--- a/pkg/operator/target_config_reconciler.go
+++ b/pkg/operator/target_config_reconciler.go
@@ -368,19 +368,19 @@ func (c TargetConfigReconciler) sync(ctx context.Context, syncCtx factory.SyncCo
 	specAnnotations["service/"+webhookService.Name] = webhookService.GetResourceVersion()
 
 	// From here, we will create our cluster wide resources.
-	err = c.manageAPIService(ownerReference)
+	err = c.manageAPIService(specAnnotations, ownerReference)
 	if err != nil {
 		klog.Error("unable to manage visibility apiservice")
 		return err
 	}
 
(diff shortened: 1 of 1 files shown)


rphillips:
The spec annotations have special magic behind them. If the resource is not changed within the operator, then the same ResourceVersion() is returned. Ultimately, the lower level code hashes all the specAnnotations, and it can detect if there is an update. If the hashes are the same, then it knows to skip the update. If the hashes are different, then it performs the update.

I don't believe there is a limit, so we should be ok there.

MaysaMacedo:
> The spec annotations have special magic behind them. If the resource is not changed within the operator, then the same ResourceVersion() is returned. Ultimately, the lower level code hashes all the specAnnotations, and it can detect if there is an update. If the hashes are the same, then it knows to skip the update. If the hashes are different, then it performs the update.
> 
> I don't believe there is a limit, so we should be ok there.

@rphillips If the `specAnnotation` changes [the Kueue deployment](https://github.com/openshift/kueue-operator/blob/main/pkg/operator/target_config_reconciler.go#L1287) would be re-created. I'm trying to understand what would be the benefit of re-creating this deployment for the specific cases of the assets highlighted in this PR. Thoughts?

I don't mind including this change. I'm just trying to understand it better.

rphillips:
No problem. Presumably, if any of the required resources have changed the operand would need to be redeployed or restarted. It will also catch 'image' changes for upgrades. If the update hashes to the same value, then there isn't an update resulting in the reconcile() function being called far less.

PR #852, opened 2025-10-22
fix kueue service monitor with correct port to scrape metrics

mukrishn:
The metrics-service added during the installation is configured with an incorrect port reference, causing Prometheus to fail in scraping metrics from the Kueue endpoint -  Updated the right port name, from [this](https://github.com/openshift/kueue-operator/blob/main/pkg/operator/target_config_reconciler.go#L1408) spec

Also the service selector label `control-plane: controller-manager` selects more than 1 service from the namespace, causes multiple invalid entries in to prometheus target config - Added a new label and updated the selector.

Fixes: [RHOAIENG-36974](https://issues.redhat.com/browse/RHOAIENG-36974)

### DIFF
--- a/pkg/operator/target_config_reconciler.go
+++ b/pkg/operator/target_config_reconciler.go
@@ -1405,7 +1405,7 @@ func (c *TargetConfigReconciler) manageServiceMonitor(ctx context.Context, kueue
 				{
 					Interval:        "30s",
 					Path:            "/metrics",
-					Port:            "https", // Name of the port you want to monitor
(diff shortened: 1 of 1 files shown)


rphillips:
We need to fix this upstream and downstream. Opened 2 PRs:

https://github.com/kubernetes-sigs/kueue/pull/7371
https://github.com/openshift/kueue-operator/pull/856

The kueue-operator PR won't work until upstream merges, and get's backported to 0.14.

PR #863, opened 2025-10-25
add linter to detect if changes occur in bindata that will be overwritten

kannon92:


kannon92:
> bindata is largely autogenerated. It should not be manually changed.

Correct. I'm trying to avoid the scenario where someone modified that in a PR and we miss it.

rphillips:
Gotcha. We should also do something like: https://github.com/openshift/kueue-operator/pull/871

Makefile
```diff
@@ -186,6 +187,10 @@ sync-manifests:
 sync-manifests-from-submodule:
 	hack/sync_manifests.py --src-dir upstream/kueue/src/config/default/
 
+.PHONY: check-sync-manifests
```
sohankunkerkar: Have you tried using Podman here? For e.g.
```sh
.PHONY: check-sync-manifests
check-sync-manifests:
	@podman run --rm \
		-v $(PWD):/workspace:Z \
		-w /workspace \
		python:3.11-slim \
		sh -c " \
			echo 'Installing dependencies...'; \
			apt-get update -qq > /dev/null 2>&1; \
			apt-get install -y -qq git curl jq > /dev/null 2>&1; \
			echo 'Fetching latest kustomize version...'; \
			KUSTOMIZE_VERSION=\$$(curl -s https://api.github.com/repos/kubernetes-sigs/kustomize/releases/latest | jq -r '.tag_name' | sed 's/kustomize\///'); \
			curl -sL https://github.com/kubernetes-sigs/kustomize/releases/download/kustomize%2Fv\$${KUSTOMIZE_VERSION}/kustomize_v\$${KUSTOMIZE_VERSION}_linux_amd64.tar.gz | tar xz -C /usr/local/bin > /dev/null 2>&1; \
			pip install pyyaml requests > /dev/null 2>&1; \
			echo 'Running check_bindata_conflicts.py...'; \
			python3 hack/check_bindata_conflicts.py \
		"
```

hack/check_bindata_conflicts.py
```diff
@@ -0,0 +1,308 @@
+#!/usr/bin/env -S uv run --with pyyaml --with requests
```
sohankunkerkar: With https://github.com/openshift/kueue-operator/pull/863/files#r2492538226, you can simply make the following change:
```suggestion
#!/usr/bin/env python3
```

Makefile
kannon92: podman isn't in the CI...

openshift really makes these things difficult.

sohankunkerkar: I think you can add `podman` [here](https://github.com/openshift/release/blob/master/ci-operator/config/openshift/kueue-operator/openshift-kueue-operator-main.yaml#L40) (use that `src` for tests) and see if it helps.

PR #875, opened 2025-10-28
OCPKUEUE-341:Split operator and operator test into two suites

cpmeadors:
Having to separate suites will make it easier to run create prow jobs for the two different test suites.  A follow up PR will be needed to update prow jobs to run operator and operand jobs separately.  Right now it is just running the operator tests so CI signal is less significant until all the operand tests are running again.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,3 +25,6 @@ go.work.sum
 # env file
 .env
 .envrc
+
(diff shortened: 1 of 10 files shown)


kannon92 (commented):
Can you remove report.xml?

Also, why was the submodule updated here?

PR #917, opened 2025-11-06
WIP: switch back to main for kueue

kannon92:


kannon92:
cc @rphillips 

Should we do this now?

PR #920, opened 2025-11-06
Add rpm deps bundle builds release 1.2

kannon92:
commit 1: fix rpm hermetic build issues
commit 2: fix cel expressions so bundle runs in CI


kannon92:
PR Group passed so I think this is good.

going to force merge to get ec working.

PR #1018, opened 2025-11-14
allow one to run tests sequentially

kannon92:
This allows one to run upstream tests without having to manually unapply the patches.


rphillips:
We might not need to revert the patches if we see if they were applied.

PR #1046, opened 2025-11-16
Exclude Flakey Tests

kannon92:
https://github.com/kubernetes-sigs/kueue/blob/main/test/e2e/singlecluster/leaderworkerset_test.go#L688C15-L688C65

This test, like others, is assuming that higher priority workloads won't schedule and lower priority workloads will be able to schedule.

Upstream makes assumptions about the number of CPUs needed in their prow cluster.

Its not entirely clear if the flakes are due to this but we disable similar tests when they rely on "1" cpu with multiple replicas.

### DIFF
--- a/upstream/kueue/e2e-test-ocp.sh
+++ b/upstream/kueue/e2e-test-ocp.sh
@@ -54,15 +54,14 @@ skips=(
         Fair
         # we do not enable this feature in our operator
         TopologyAwareScheduling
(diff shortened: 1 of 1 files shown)


rphillips:
Failed on 'Kueue when Creating a Job With Queueing [It] Should unsuspend a job and set nodeSelectors'

kannon92:
@rphillips maybe we can merge this and iterate on flakes?

PR #1066, opened 2025-11-18
OCPKUEUE-365: Add e2e for labelPolicy=None with JobSet

MaysaMacedo:
This commit adds a new e2e scenario that checks
if when the `labelPolicy` is None, the JobSets created
in namespaces labeled with the openshift managed are suspened
and JobSets that are created in unlabeled namespaces
are not suspened and not managed by Kueue.


test/e2e/e2e_suite_test.go
```diff
@@ -64,6 +66,7 @@ func TestE2E(t *testing.T) {
 }
 
 var _ = BeforeSuite(func() {
+	log.SetLogger(zap.New(zap.UseDevMode(true)))
```
kannon92: Why is this change here?

go.mod
```diff
@@ -29,6 +29,7 @@ require (
 	k8s.io/utils v0.0.0-20250321185631-1f6e0b77f77e
 	sigs.k8s.io/controller-runtime v0.20.4
 	sigs.k8s.io/controller-tools v0.4.1
+	sigs.k8s.io/jobset v0.7.1
```
kannon92: This is a really old version of jobset.

Could you use v0.10?

kannon92:
Try a rebase when you get a chance.

rphillips:
The error in the 4-20 suite looks like the same race that https://github.com/openshift/kueue-operator/pull/1141 is trying to fix. 1141 isn't done yet, but I plan on looking at it on Monday.

PR #1070, opened 2025-11-18
change back to upstream main

rphillips:


kannon92:
/lgtm
/approve

We expect this CI to fail because bundle still picks old operand image.

The failures will happen and we will get new nudges. After that, we expect CI to stablize.

PR #1087, opened 2025-11-20
exclude node selector job

kannon92:


rphillips:
I'm not sure. I can get these to reliably pass locally.

rphillips:
/hold

I think we should investigate this a bit further.

kannon92:
> /hold
> 
> I think we should investigate this a bit further.

I have spent a significant time investigating this failure.

The must gathers all point to difficulties scheduling this job. Looking at kube-scheduler, when this test flakes, kube-scheduler has challenges scheduling the pods. 


I am fine to investigate more on main but I think for release-1.2 we should exclude this job.

https://github.com/openshift/kueue-operator/pull/1112

PR #1131, opened 2025-11-25
fix: Graceful shutdown with context propagation and network policy fixes

rphillips:
This PR fixes critical issues with graceful shutdown, network policies, and e2e test race conditions in the Kueue operator.

  Commits

  1. networkpolicy: fixes for egress being blocked (946a58c)
  - Renamed network policies with numbered prefixes (10-, 99-) to control application order
  - Added explicit egress rules to allow DNS lookups, API server access, and metrics collection
  - Fixed egress blocking issues that prevented the operator from communicating with external services

  2. fix sigterm handling by passing the context (793c3df)
  - Removed stored context (c.ctx) from TargetConfigReconciler struct
  - Refactored all reconciliation functions to accept and propagate cancellable context parameter
  - Ensures SIGTERM signals properly cancel ongoing API operations for graceful shutdown
  - Enables proper leader election lease release, reducing new pod startup delays from ~3 minutes to seconds

  3. allow workloads that might be finished (5725e6c)
  - Fixed e2e test race conditions where workloads complete before tests verify admission status
  - Modified verifyWorkloadCreated to accept workloads in either "admitted" or "finished" state
  - Added explicit waits for all workloads to be created before checking pending workload counts
  - Prevents false test failures when jobs complete faster than test verification

  Impact

  - Graceful Shutdown: Operator now properly cancels API operations on SIGTERM, enabling clean pod termination within the 30-second grace period
  - Leader Election: New pods can acquire leader lease immediately instead of waiting 137+ seconds
  - Network Connectivity: Fixed egress blocking that prevented metrics, visibility API, and webhook traffic
  - Test Reliability: E2e tests are now resilient to timing variations and workload completion races

### DIFF
--- a/test/e2e/e2e_operator_test.go
+++ b/test/e2e/e2e_operator_test.go
@@ -23,6 +23,7 @@ import (
 	"fmt"
 	"slices"
 	"strings"
+	"time"
 
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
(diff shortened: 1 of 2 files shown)


rphillips:
Yeah, will do.

PR #1134, opened 2025-11-26
[release-1.2] fix: Graceful shutdown with context propagation and network policy fixes

rphillips:
Pick of https://github.com/openshift/kueue-operator/pull/1131

### DIFF
--- a/pkg/operator/target_config_reconciler.go
+++ b/pkg/operator/target_config_reconciler.go
@@ -425,28 +425,6 @@ func (c *TargetConfigReconciler) sync(ctx context.Context, syncCtx factory.SyncC
 	}
 	specAnnotations["rolebinding/"+roleBindingLeader.Name] = hash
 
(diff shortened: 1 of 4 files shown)


pkg/operator/target_config_reconciler.go
```diff
@@ -423,8 +425,30 @@ func (c *TargetConfigReconciler) sync(ctx context.Context, syncCtx factory.SyncC
 	}
 	specAnnotations["rolebinding/"+roleBindingLeader.Name] = hash
 
+	managerSecretsRole, _, err := c.manageRole(ctx, "assets/kueue-operator/role-manager-secrets.yaml", ownerReference)
```
kannon92: /hold

This was not in kueue 1.2. (0.14).

This was added in 0.15.

kannon92: I think the CI is going to fail here because these assets are not present.

PR #1177, opened 2025-12-03
add command for test coverage

cpmeadors:
PoC for analyzing test coverage and making it easier to understand.  It just runs unit tests and looks for coverage files anywhere and then spits out a summary with recommendations.  Coverage reports should probably be centralized.  Also, other test suites should be run but they need more complex setup.

Test with claude: /coverage-summary


kannon92 (commented):
/lgtm

PR #1213, opened 2025-12-12
 Add Support for External Framework field in MultiKueue

khrm:
This adds support for External Framework needed for MultiKueue.
    It is needed for supporting OpenShift Pipelines MultiCluster.

### DIFF
--- a/pkg/configmap/configmap.go
+++ b/pkg/configmap/configmap.go
@@ -31,8 +31,8 @@ import (
 	kueue "github.com/openshift/kueue-operator/pkg/apis/kueueoperator/v1"
 )
 
-func BuildConfigMap(namespace string, kueueCfg kueue.KueueConfiguration) (*corev1.ConfigMap, error) {
-	config := defaultKueueConfigurationTemplate(namespace, kueueCfg)
+func BuildConfigMap(namespace string, kueueCfg kueue.KueueConfiguration, gvrToKind map[string]string) (*corev1.ConfigMap, error) {
+	config := defaultKueueConfigurationTemplate(namespace, kueueCfg, gvrToKind)
 	cfg, err := yaml.Marshal(config)
 	if err != nil {
 		return nil, err
@@ -100,20 +100,24 @@ func buildLabelKeysCopy(labelKeys []kueue.LabelKeys) []string {
 	return ret
 }
 
-func mapOperatorMultiKueueToKueue(multiKueue *kueue.MultiKueue) *configapi.MultiKueue {
+func mapOperatorMultiKueueToKueue(multiKueue *kueue.MultiKueue, gvrToKind map[string]string) *configapi.MultiKueue {
 	if multiKueue == nil {
 		return nil
 	}
 	return &configapi.MultiKueue{
-		ExternalFrameworks: buildMultiKueueExternalFrameworkList(multiKueue.ExternalFrameworks),
+		ExternalFrameworks: buildMultiKueueExternalFrameworkList(multiKueue.ExternalFrameworks, gvrToKind),
 	}
 }
 
-func buildMultiKueueExternalFrameworkList(kueuelist []kueue.ExternalFramework) []configapi.MultiKueueExternalFramework {
+func buildMultiKueueExternalFrameworkList(kueuelist []kueue.ExternalFramework, gvrToKind map[string]string) []configapi.MultiKueueExternalFramework {
 	ret := []configapi.MultiKueueExternalFramework{}
 	for _, val := range kueuelist {
+		kind := val.Resource
+		if k, ok := gvrToKind[fmt.Sprintf("%s/%s/%s", val.Group, val.Version, val.Resource)]; ok {
+			kind = k
+		}
 		ret = append(ret, configapi.MultiKueueExternalFramework{
-			Name: fmt.Sprintf("%s.%s.%s", val.Resource, val.Version, val.Group),
+			Name: fmt.Sprintf("%s.%s.%s", kind, val.Version, val.Group),
 		})
 	}
 	return ret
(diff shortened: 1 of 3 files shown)


pkg/apis/kueueoperator/v1/types.go
```diff
@@ -70,6 +70,12 @@ type KueueConfiguration struct {
 	// This default could change over time.
 	// +optional
 	Preemption Preemption `json:"preemption"`
+	// multiKueue controls the behaviour of the MultiKueue AdmissionCheck Controller.
```
kannon92: So I was debating about two approaches.

1) Add API support for this for MK.
2) Automatically set up MK settings based on external framework fields.

Curious on your thoughts here?

Is there any reason why you want to configure external frameworks for tekton-kueue for single cluster but not have it enabled in case you are working in MK environment?

khrm:
Verified:
```
➜  oc logs -f deployment/kueue-controller-manager  -n openshift-kueue-operator  | grep tekton.dev
Found 2 pods, using pod/kueue-controller-manager-79f58cdb4d-g5qmk
{"level":"info","ts":"2025-12-16T19:36:28.434538603Z","logger":"setup","caller":"kueue/main.go:529","msg":"Successfully loaded configuration","config":"apiVersion: config.kueue.x-k8s.io/v1beta2\nclientConnection:\n  burst: 100\n  qps: 50\ncontroller:\n  groupKindConcurrency:\n    ClusterQueue.kueue.x-k8s.io: 1\n    Job.batch: 5\n    LocalQueue.kueue.x-k8s.io: 1\n    Pod: 5\n    ResourceFlavor.kueue.x-k8s.io: 1\n    Workload.kueue.x-k8s.io: 5\nhealth:\n  healthProbeBindAddress: :8081\nintegrations:\n  externalFrameworks:\n  - pipelinerun.v1.tekton.dev\n  frameworks:\n  - batch/job\n  - pod\n  - deployment\n  - statefulset\n  - jobset.x-k8s.io/jobset\n  - leaderworkerset.x-k8s.io/leaderworkerset\ninternalCertManagement:\n  enable: false\nkind: Configuration\nleaderElection:\n  leaderElect: true\n  leaseDuration: 2m17s\n  renewDeadline: 1m47s\n  resourceLock: leases\n  resourceName: c1f6bfd2.kueue.x-k8s.io\n  resourceNamespace: \"\"\n  retryPeriod: 26s\nmanageJobsWithoutQueueName: false\nmanagedJobsNamespaceSelector:\n  matchLabels:\n    kueue.openshift.io/managed: \"true\"\nmetrics:\n  bindAddress: :8443\n  enableClusterQueueResources: true\nmultiKueue:\n  dispatcherName: kueue.x-k8s.io/multikueue-dispatcher-all-at-once\n  externalFrameworks:\n  - name: pipelinerun.v1.tekton.dev\n  gcInterval: 1m0s\n  origin: multikueue\n  workerLostTimeout: 15m0s\nnamespace: openshift-kueue-operator\nwebhook:\n  certDir: /tmp/k8s-webhook-server/serving-certs\n  port: 9443\n"}
{"level":"info","ts":"2025-12-16T19:36:28.471469767Z","logger":"setup","caller":"kueue/main.go:390","msg":"Creating external framework MultiKueue adapter","gvk":"tekton.dev/v1, Kind=pipelinerun"}
```

pkg/apis/kueueoperator/v1/types.go
kannon92: I think the API support is fine.

It matches upstream and we reuse the same type so I don't think the API is too much to add.

pkg/apis/kueueoperator/v1/types.go
```diff
@@ -320,3 +326,18 @@ type Preemption struct {
 	// +required
 	PreemptionPolicy PreemptionPolicy `json:"preemptionPolicy"`
 }
+
+// MultiKueue controls the behaviour of the MultiKueue AdmissionCheck Controller.
+type MultiKueue struct {
+	// externalFrameworks are a list of GroupVersionKinds that should be supported
```
JoelSpeed: Should use GVR to be precise, GVK can be ambiguous. Suggestion would be to copy the existing API patterns and create a structure GVR list rather than a string here

khrm: Can gvk be ambigous if it's not subresource? BTW, it's gvk upstream. We need gvk here, otherwise it's, not working properly. It doesn't initialize proper adapter.

khrm: Can gvk be ambigous if it's not subresource? BTW, it's gvk upstream. We need gvk here, otherwise it's, not working properly. It doesn't initialize proper adapter.

khrm: Can gvk be ambigous if it's not subresource? BTW, it's gvk upstream. We need gvk here, otherwise it's, not working properly. It doesn't initialize proper adapter.

JoelSpeed: Yes, GVK can be ambiguous even when it's not a subresource, please review https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#kind-vs-resource

if Kueue upstream is using GVK, that's fine, it's not hard to map GVR to GVK, we could translate within our abstraction. We should however promote the Kueue APIs as they evolve to move to GVR for the reasons outlined in the conventions docs

khrm:
I have changed this to fetch mapping of GVR to GVK and apply that.

pkg/apis/kueueoperator/v1/types.go
kannon92: I believe you added some code to fix this?

pkg/apis/kueueoperator/v1/types.go
```diff
@@ -320,3 +326,18 @@ type Preemption struct {
...
+	// by the generic MultiKueue adapter. Each entry defines how to handle a specific
+	// GroupVersionKind (GVK) for MultiKueue operations.
+	// externalFrameworks are optional and should only be used if you have external frameworks
+	// that need MultiKueue support.
+	// externalFrameworks, if specified, can not have more than 32 items.
+	// +listType=map
+	// +listMapKey=group
+	// +kubebuilder:validation:MaxItems=32
```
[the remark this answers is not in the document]
kannon92: This is done.

kannon92:
:(

I believe we have something going wrong in our CI. I have #1301 or #1302 to hopefully fix the CI.

PR #1215, opened 2025-12-12
remove kueue go.mod replacement and use pinned upstream

kannon92:


MaysaMacedo (approved):
/lgtm

kannon92:
> This might not be safe. We want the go.mod to bring in versioned Kueue APIs for the Operator/Tests to use.

Wouldn't they be versioned based on whatever our submodule is on?

We could just use upstream repo but there could be a chance that the APIs we use don't match downstream.

rphillips:
/retitle remove kueue go.mod pin and use pinned upstream

rphillips:
/retitle remove kueue go.mod replacement and use pinned upstream

PR #1685, opened 2026-04-07
Add support to Admission Fair Sharing API

MaysaMacedo:
This PR adds the support for the `AdmissionFairSharing` [configs](https://github.com/kubernetes-sigs/kueue/blob/15764bcf9d471f7452697a43df09c7bbf20c888a/apis/config/v1beta2/configuration_types.go#L598-L611) in the kueue-operator API Spec.

The type of ResourceWeights was modified from `map[corev1.ResourceName]float64` to `map[corev1.ResourceName]string` to address the following warning:
```
go run ./vendor/sigs.k8s.io/controller-tools/cmd/controller-gen crd paths=./pkg/apis/kueueoperator/v1/... output:crd:dir=./manifests                                     
/home/mdemaced/workspace/openshift/kueue-operator/pkg/apis/kueueoperator/v1/types.go:158:42: found float, the usage of which is highly discouraged, as support for them v
aries across languages. Please consider serializing your float as string instead. If you are really sure you want to use them, re-run with crd:allowDangerousTypes=true  
Error: not all generators ran successfully
```
Internally that type is converted back to `float64` to be accepted by the Kueue API.

**Special notes for your reviewer:**

Assisted-by:  Claude


pkg/configmap/configmap.go
```diff
@@ -211,7 +215,30 @@ func buildFeatureGates(resources kueue.Resources, frameworks []kueue.KueueIntegr
...
+		f, err := strconv.ParseFloat(string(entry.Weight), 64)
+		if err != nil {
+			return nil, fmt.Errorf("failed to parse resource weight for resource %s: %w", entry.Name, err)
+		}
+		resourceWeights[corev1.ResourceName(entry.Name)] = f
+	}
+	return &configapi.AdmissionFairSharing{
+		UsageHalfLifeTime:     v1.Duration{Duration: time.Duration(ptr.Deref(admissionFairSharing.UsageHalfLifeTimeSeconds, 0)) * time.Second},
+		UsageSamplingInterval: v1.Duration{Duration: time.Duration(ptr.Deref(admissionFairSharing.UsageSamplingIntervalSeconds, 0)) * time.Second},
+		ResourceWeights:       resourceWeights,
```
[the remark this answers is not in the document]
MaysaMacedo: In Kueue, when `UsageHalfLifeTime` is not defined in the configMap, the kueue-controller-manager generates the config as `usageHalfLifeTime: 0s`, so we're safe.

pkg/apis/kueueoperator/v1/types.go
```diff
@@ -82,6 +82,13 @@ type KueueConfiguration struct {
 	// If multiKueue is not specified, MultiKueue is disabled.
 	// +optional
 	MultiKueue *MultiKueue `json:"multiKueue,omitempty"`
+
+	// admissionFairSharing enables fair sharing at admission time.
+	// It makes Kueue order pending workloads by each LocalQueue's accumulated resource usage,
+	// so that queues which have consumed fewer resources are admitted first.
+	// This section configures the usage decay rate, the sampling frequency, and per-resource weights.
+	// +optional
+	AdmissionFairSharing *AdmissionFairSharing `json:"admissionFairSharing,omitempty"`
```
everettraven: The underlying type has no required fields, making `admissionFairSharing: {}` currently an accepted input.

What does it mean to set `admissionFairSharing: {}` as opposed to omitting the field altogether?

pkg/apis/kueueoperator/v1/types.go
```diff
@@ -128,6 +135,50 @@ const (
 	KueueIntegrationSparkApplication KueueIntegration = "SparkApplication"
 )
 
+// AdmissionFairSharing provides configuration of intervals
+// and resource weights for Admission Fair Sharing.
+type AdmissionFairSharing struct {
+	// usageHalfLifeTimeSeconds indicates the time in seconds after which the current usage will decrease by a half.
+	// If set to 0, usage will be reset to 0 immediately.
+	// +kubebuilder:validation:Minimum=0
```
everettraven: Note: Generated documentation doesn't include these markers. We recommend adding explicit text to the GoDoc describing the constraints enforced by the markers so they show up in the generated documentation that end users see when doing a `kubectl explain ...`.

pkg/apis/kueueoperator/v1/types.go
```diff
@@ -128,6 +135,50 @@ const (
...
+	// +optional
+	UsageHalfLifeTimeSeconds *int32 `json:"usageHalfLifeTimeSeconds,omitempty"`
```
everettraven: Is there a reasonable upper bound you can enforce here?

Leaving this wide open means that someone can set this to something like `usageHalfLifeTimeSeconds: 2147483647`, which is equivalent to ~68 years. Is that a reasonable value to set?

pkg/apis/kueueoperator/v1/types.go
```diff
@@ -128,6 +135,50 @@ const (
 	KueueIntegrationSparkApplication KueueIntegration = "SparkApplication"
 )
 
+// AdmissionFairSharing provides configuration of intervals
+// and resource weights for Admission Fair Sharing.
+type AdmissionFairSharing struct {
+	// usageHalfLifeTimeSeconds indicates the time in seconds after which the current usage will decrease by a half.
+	// If set to 0, usage will be reset to 0 immediately.
+	// +kubebuilder:validation:Minimum=0
+	// +optional
```
everettraven, line 174: What happens if I don't set this field?

pkg/apis/kueueoperator/v1/types.go
```diff
@@ -128,6 +135,50 @@ const (
...
+	UsageHalfLifeTimeSeconds *int32 `json:"usageHalfLifeTimeSeconds,omitempty"`
+
+	// usageSamplingIntervalSeconds is the frequency in seconds that Kueue updates consumedResources in FairSharingStatus.
+	// When omitted or set to 0, kueue will decide the default, which is subject to change over time.
```
everettraven: If omission and explicitly set to `0` mean the same thing, make `0` an invalid value. We strongly encourage only having a single way for a user to "say" the same thing to prevent confusion over what is the correct way to configure something.

By doing this, this field also no longer needs to be a pointer.

pkg/apis/kueueoperator/v1/types.go
```diff
@@ -128,6 +135,50 @@ const (
...
+	// The current default is 300 (5 minutes).
+	// +kubebuilder:validation:Minimum=0
+	// +optional
+	UsageSamplingIntervalSeconds *int32 `json:"usageSamplingIntervalSeconds,omitempty"`
```
everettraven: Similar comment regarding a reasonable upper bounds.

pkg/apis/kueueoperator/v1/types.go
```diff
@@ -128,6 +135,50 @@ const (
...
+
+	// resourceWeights assigns weights to resources which are then used to calculate LocalQueue's
+	// resource usage and order Workloads.
+	// When omitted, kueue will decide the default, which is subject to change over time.
+	// The current default weight is 1 for any resource.
+	// +listType=map
+	// +listMapKey=name
+	// +kubebuilder:validation:MaxItems=16
+	// +optional
+	ResourceWeights []ResourceWeight `json:"resourceWeights,omitempty"`
```
everettraven, line 197: Add a minimum items constraint of 1 to prevent unstructured clients from being able to set `resourceWeights: []`.

pkg/apis/kueueoperator/v1/types.go
```diff
@@ -128,6 +135,50 @@ const (
...
+	// +optional
+	UsageSamplingIntervalSeconds *int32 `json:"usageSamplingIntervalSeconds,omitempty"`
+
+	// resourceWeights assigns weights to resources which are then used to calculate LocalQueue's
+	// resource usage and order Workloads.
+	// When omitted, kueue will decide the default, which is subject to change over time.
+	// The current default weight is 1 for any resource.
+	// +listType=map
+	// +listMapKey=name
+	// +kubebuilder:validation:MaxItems=16
```
everettraven, line 195: Document the constraints enforced by these markers explicitly in the GoDoc please.

pkg/apis/kueueoperator/v1/types.go
```diff
@@ -128,6 +135,50 @@ const (
...
+	// +optional
+	ResourceWeights []ResourceWeight `json:"resourceWeights,omitempty"`
+}
+
+// ResourceWeight associates a resource name with a weight used for fair sharing calculations.
+type ResourceWeight struct {
+	// name is the Kubernetes resource name (e.g., "cpu", "memory", "nvidia.com/gpu").
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:MaxLength=253
+	// +required
```
everettraven: If this is a Kubernetes resource name, we should probably be matching the validation of https://github.com/kubernetes/kubernetes/blob/0275382260bc0d5279b703d21ec7ad4db35b8d78/staging/src/k8s.io/apimachinery/pkg/api/validate/strfmt.go#L278-L285

IIRC this is the same validation that is performed on Pod resource request names.

pkg/apis/kueueoperator/v1/types.go
```diff
@@ -128,6 +135,50 @@ const (
...
+	Name string `json:"name,omitempty"`
+
+	// weight is a string representation of a non-negative float64 value
+	// used as a weight for a resource in fair sharing calculations (e.g., "1.0", "0.5").
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:MaxLength=32
```
everettraven, line 216: 32 feels a bit high to me - would setting something like `weight: "0.00000000000000000000000000001"` be reasonable here?

pkg/apis/kueueoperator/v1/types.go
kannon92: I did hear something at KubeCon who was interested in setting really large Admission Fair sharing as a way to mimic budgets running out.

pkg/apis/kueueoperator/v1/types.go
MaysaMacedo: It means the default values would be enforced. For example, when I set the `admissionFairSharing: {}` in the kueue configuration  I get the following enforced in Kueue: `"admissionFairSharing:\n  usageHalfLifeTime: 0s\n  usageSamplingInterval: 5m0s`

pkg/apis/kueueoperator/v1/types.go
MaysaMacedo: The Kueue controller would load it as `usageHalfLifeTime: 0s`. This probably happens because the type is of Duration https://github.com/kubernetes-sigs/kueue/blob/main/apis/config/v1beta2/configuration_types.go#L601

pkg/apis/kueueoperator/v1/types.go
MaysaMacedo: ack, I also modified UsageHalfLifeTimeSeconds to follow the same pattern.

pkg/apis/kueueoperator/v1/types.go
MaysaMacedo: This field would correspond to this one https://github.com/kubernetes-sigs/kueue/blob/e278cc9285c8caae4931430031547215c6b1252a/vendor/k8s.io/api/core/v1/types.go#L6934

I also see this on the docs https://github.com/kubernetes-sigs/kueue/blob/e278cc9285c8caae4931430031547215c6b1252a/vendor/k8s.io/api/core/v1/types.go#L6936-L6940
so I wonder which one to follow, the one you mentioned or the this one I'm linking?

pkg/apis/kueueoperator/v1/types.go
MaysaMacedo: I'm not certain on this. I need to investigate more.

pkg/apis/kueueoperator/v1/types.go
everettraven: I would go with the source of truth for Kueue, so the one you are linking. We should make sure that all of those constraints are being appropriately enforced at admission time by validation markers and that they are explicitly documented in the GoDoc.

pkg/apis/kueueoperator/v1/types.go
everettraven: What happens if I just omit the field altogether?

pkg/apis/kueueoperator/v1/types.go
kannon92: https://github.com/kubernetes-sigs/kueue/issues/10434

Very similar issue filed today on this feature.

pkg/apis/kueueoperator/v1/types.go
everettraven: GitHub won't let me resolve this comment. Please feel free to resolve it.

pkg/apis/kueueoperator/v1/types.go
everettraven: Feel free to resolve this comment. GitHub won't let me.

pkg/apis/kueueoperator/v1/types.go
everettraven: Feel free to resolve this comment. GitHub won't let me.

pkg/apis/kueueoperator/v1/types.go
MaysaMacedo: No `admissionFairSharing` is set in the configMap and in the kueue-controller-manager.

everettraven: Thanks for clarifying. I've got a couple thoughts here.

1. If we can clarify the GoDoc as to this behavior, that would likely be useful to end users as it was also helpful to me.

2. We typically try to avoid having "nothing" (i.e the empty value) mean something because it can be a bit confusing for end users that are less familiar with the API. What do you think of something like:
```yaml
admissionFairSharing:
  configuration: Default
```
to represent explicitly choosing the default configuration parameters and something like:
```yaml
admissionFairSharing:
  configuration: Custom
  custom:
    # existing admission fair sharing fields for customized configuration. At least 1 must be explicitly set.
```
to represent explicitly choosing to customize the configuration parameters?

Just a note, `configuration` is just an arbitrary name I came up with - if you take this approach feel free to rename to whatever feels the most appropriate.

kannon92: @everettraven would it be better to make AdmissionFairSharing a non pointer?

A lot of our other APIs are not pointers.

everettraven: Pointer/non-pointer depends on the representation that you need.

Maintaining it as-is where omission means that admission fair sharing is not configured and setting to the zero-value (`{}`) means to apply the defaults, you need this to be a pointer.

Taking the alternative approach I suggested, you could probably get away with this being a non-pointer

MaysaMacedo: FYI: with the latest code in the PR, I added a log locally just to check out what the admissionFairSharing would look like internally in the kueue-operator spec when it's not specified and this is what I get:
`{Configuration: Custom:{UsageHalfLifeTimeSeconds:0 UsageSamplingIntervalSeconds:0 ResourceWeights:[]}}` which is skipped [here](https://github.com/openshift/kueue-operator/pull/1685/changes#diff-6aa7780e80409d3ad0fb397be31e6f2d64ab520750d4317267f7138ebcee6606R227).
I wonder if that is ok

pkg/apis/kueueoperator/v1/types.go
```diff
@@ -128,6 +135,86 @@ const (
...
+	// +optional
+	ResourceWeights []ResourceWeight `json:"resourceWeights,omitempty"`
+}
+
+// ResourceWeight associates a resource name with a weight used for fair sharing calculations.
+type ResourceWeight struct {
+	// name is the Kubernetes resource name (e.g., "cpu", "memory", "nvidia.com/gpu").
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:MaxLength=253
+	// +kubebuilder:validation:XValidation:rule="!format.qualifiedName().validate(self).hasValue()",message="must be a qualified name consisting of alphanumeric characters, hyphens, underscores, dots, with an optional DNS subdomain prefix and forward slash (e.g., 'nvidia.com/gpu' or 'cpu')"
```
everettraven: Explicitly document these constraints in the GoDoc please :)

pkg/apis/kueueoperator/v1/types.go
everettraven: Feel free to resolve this thread. GitHub won't let me.

pkg/apis/kueueoperator/v1/types.go
everettraven: Just a note for once this is decided on - please make sure to include any constraints explicitly in the GoDoc.

pkg/apis/kueueoperator/v1/types.go
everettraven: That seems fine to me

pkg/apis/kueueoperator/v1/types.go
MaysaMacedo: FYI, I used the validation from here https://github.com/kubernetes-sigs/kueue/blob/2438fa5f939120dd2e43111c0d078d7f17e63b29/vendor/k8s.io/apimachinery/pkg/api/validate/content/kube.go#L43-L72

pkg/apis/kueueoperator/v1/types.go
MaysaMacedo: I will update this GoDoc when there is a conclusion on upstream, for now I propose that we leave as is. WDYT?

MaysaMacedo:
@kannon92 @everettraven This PR is ready for another look. Could you check it when possible? Thanks!

kannon92:
> **Note:** This review was AI-generated using Claude.

## PR Review: Add support to Admission Fair Sharing API

**+660 / -3** across 11 files

### Overview

This PR adds `AdmissionFairSharing` configuration to the kueue-operator API, allowing users to enable Kueue's admission-time fair sharing feature. The design uses a discriminated union pattern (`Default`/`Custom`) consistent with the existing `GangScheduling` API. The `Custom` variant exposes `usageHalfLifeTimeSeconds`, `usageSamplingIntervalSeconds`, and `resourceWeights`. Resource weights are modeled as `[]ResourceWeight` (list-type map) with string-serialized floats to avoid controller-gen's float restriction.

### Positives

- **Good API design** — The `Default`/`Custom` union with CEL validation (`custom is required when configuration is Custom, and forbidden otherwise`) is clean and consistent with the existing `GangScheduling` pattern.
- **Strong CRD validation** — CEL rules on `resourceWeights` values (`^[0-9]+(\.[0-9]+)?$`), `format.qualifiedName()` on resource names, min/max bounds on integer seconds fields, `minProperties: 1` on custom, and list-type map with max 16 items.
- **Reasonable test coverage** for the configmap builder with the `resourceWeights` case.

### Issues

#### 1. Zero-value durations emitted when only `resourceWeights` is set (Medium)

`pkg/configmap/configmap.go:233-234` — When only `resourceWeights` is set on `Custom`, the int32 fields default to `0`, producing:
```yaml
usageHalfLifeTime: 0s
usageSamplingInterval: 0s
```

This is visible in the test expectation at `configmap_test.go:879-880`. A `0s` half-life has specific semantics ("usage reset immediately") which may be fine, but `usageSamplingInterval: 0s` means "sample every 0 seconds" — which likely isn't what Kueue expects and bypasses its default of 300s. The builder should omit these fields when they're 0 so Kueue can apply its own defaults. Consider:

```go
if custom.UsageHalfLifeTimeSeconds > 0 {
    result.UsageHalfLifeTime = v1.Duration{Duration: time.Duration(custom.UsageHalfLifeTimeSeconds) * time.Second}
}
if custom.UsageSamplingIntervalSeconds > 0 {
    result.UsageSamplingInterval = v1.Duration{Duration: time.Duration(custom.UsageSamplingIntervalSeconds) * time.Second}
}
```

Or use pointer types in `AdmissionFairSharingCustom` (`*int32`) to distinguish "not set" from "explicitly zero" — though zero is forbidden by the `Minimum=1` constraint anyway, so the omit-on-zero approach is simpler.

#### 2. Missing test for `Default` configuration (Low-Medium)

There's no test for `AdmissionFairSharingConfigurationDefault`. This should produce an `AdmissionFairSharing` result with no custom fields — worth verifying the configmap output is correct (just `admissionFairSharing: {}` or similar).

#### 3. Missing test for empty/unset `AdmissionFairSharing` (Low)

There's no explicit test that when `AdmissionFairSharing` is unset (zero value), the output configmap has no `admissionFairSharing` key at all. The existing tests may cover this implicitly, but an explicit case would prevent regressions.

#### 4. `AdmissionFairSharing` is a value type, not a pointer (Low)

`pkg/apis/kueueoperator/v1/types.go:91` — `AdmissionFairSharing` is embedded as a value (`AdmissionFairSharing AdmissionFairSharing`) with `omitzero`, while the adjacent `MultiKueue` uses a pointer with `omitempty`. The `omitzero` approach works for Go serialization, but it differs from the pattern used elsewhere in this struct. This is likely intentional given `omitzero` is the newer convention — just flagging for consistency awareness.

#### 5. Weight regex doesn't accept scientific notation or leading `+` (Nit)

The CEL regex `^[0-9]+(\.[0-9]+)?$` rejects values like `1e-3`, `+0.5`, or `.5`. The upstream Kueue API accepts `float64`, so these are valid floats. This is a conscious restriction (keeps it simple), but worth documenting that only decimal notation is accepted.

#### 6. `buildAdmissionFairSharing` returns empty struct for `Default` (Low)

`configmap.go:230` — When `Configuration == "Default"`, the function returns `&configapi.AdmissionFairSharing{}` (empty struct). Verify that Kueue's config parser treats an empty `AdmissionFairSharing` object the same as "enable with defaults." If Kueue requires the `Enable` field or similar to be set, this would silently not enable the feature.

### Summary

The API design and validation are solid. The main concern is **issue #1** — zero-value durations being emitted when custom fields are unset, which could override Kueue's built-in defaults. I'd also add test cases for the `Default` configuration path and the unset/empty path. The rest are minor observations.

pkg/configmap/configmap_test.go
```diff
@@ -848,6 +848,72 @@ metrics:
...
   enableClusterQueueResources: true
 namespace: test
+webhook:
+  port: 9443
+`,
+				},
+			},
+			wantErr: nil,
+		},
+		"Admission Fair Sharing with only resourceWeights set": {
```
kannon92, line 915: I left an API review but I noticed some missing test cases here. We are only testing Custom here. Please add the other test cases I lay out in the review.

MaysaMacedo: point 1.  2. 5. were fixed. Note that when using the Default configuration even if the field is set as empty, the default value from Duration type, which is 0, will be enforced. Internally Kueue will modify usagesamplinginterval to 5.
For 3. the other tests already covers this.
Not sure if 6. is necessary given the default values are defined internally in Kueue during runtime.

kannon92: Can we add a test case here for Default?

MaysaMacedo: Sure, I added this one "Admission Fair Sharing with Default configuration".
Let me what do you think

pkg/apis/kueueoperator/v1/types.go
```diff
@@ -128,6 +135,90 @@ const (
...
+type AdmissionFairSharingConfiguration string
+
+const (
+	AdmissionFairSharingConfigurationDefault AdmissionFairSharingConfiguration = "Default"
+	AdmissionFairSharingConfigurationCustom  AdmissionFairSharingConfiguration = "Custom"
+)
+
+// AdmissionFairSharing provides configuration of intervals
+// and resource weights for Admission Fair Sharing.
+// +kubebuilder:validation:XValidation:rule="has(self.configuration) && self.configuration == 'Custom' ? has(self.custom) : !has(self.custom)",message="custom is required when configuration is Custom, and forbidden otherwise"
```
kannon92, line 148: Could we look into ways to validate CEL expressions?

Right now this code is untested.

pkg/apis/kueueoperator/v1/types.go
```diff
@@ -128,6 +135,90 @@ const (
...
+
+// ResourceWeight associates a resource name with a weight used for fair sharing calculations.
+type ResourceWeight struct {
+	// name consists of an optional DNS subdomain prefix (lowercase alphanumeric, hyphens, dots,
+	// up to 253 characters), followed by a slash '/', and a name segment.
+	// The name segment must be at most 63 characters, consisting of alphanumeric characters (upper or lower case),
+	// with '-', '_', and '.' allowed anywhere except the first or last character.
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:MaxLength=317
+	// +kubebuilder:validation:XValidation:rule="!format.qualifiedName().validate(self).hasValue()",message="must be a qualified name (e.g., 'nvidia.com/gpu' or 'cpu')"
```
kannon92, line 208: Let's figure out how to get testing on these CEL rules.

pkg/apis/kueueoperator/v1/types.go
everettraven: Do you anticipate the maximum length constraint being further restricted?

If so, that would be a breaking change to writers of the API - which IIRC is end-users. We should be cautious of making these kinds of breaking changes. In general, it is easier to justify increasing the maximum length than decreasing it.

pkg/apis/kueueoperator/v1/types.go
everettraven: For context, how we generally do it in openshift/api is:
1. Run envtest to stand up a lightweight Kube API server
2. Apply _only_ the generated CRD
3. Do Create/Update requests of the CR with various values and resulting output expectations

everettraven:
Aside from the maximum length constraint on the `weight` field being up in the air, the API structure LGTM.

I won't block on that and will defer to you folks on the level of risk of it being changed in the future, but you should take into account that it is generally easier to increase a maximum constraint than it is to decrease it.

kannon92:
/approve
/lgtm

/hold

I don't see anything here that can't be done as a followup. If you want to address comments on this PR feel free.

Otherwise you can unhold.

pkg/apis/kueueoperator/v1/types.go
kannon92: I'm fine if we do this as a follow up @MaysaMacedo.

pkg/apis/kueueoperator/v1/types.go
kannon92: I am not concerned with us ever restricting this. 32 seems to be a resonable value as this would be the number of pod spec requests. This seems pretty high now but if we need to increase we can.

MaysaMacedo:
/hold cancel
thanks everyone for the reviews!

pkg/apis/kueueoperator/v1/types.go
MaysaMacedo: will work on it in a follow-up PR

PR #1795, opened 2026-04-20
OCPKUEUE-613 - Add DRA e2e tests for Structured Parameters

anahas-redhat:
**Enhance the DRA Structured Parameters test suite with new scenarios:**
- DRA quota admission for Pod, Deployment, and StatefulSet workload types
- Preemption of low-priority DRA workloads by high-priority ones
- Gang scheduling enforcement for multi-pod DRA workloads
- End-to-end validation using the vendor-neutral dra-example-driver

Uses Gomega g callback pattern for better error reporting on failure. The dra-example-driver Helm installation is handled by a dedicated CI step rather than inline in the test code.

Also includes test infrastructure improvements: GenerateName for namespaces, DeferCleanup for config restore, hoisted kueueClient, WithDRAResource helper on ClusterQueueWrapper, ResourceClaim allocation verification, and standardized job commands via newDRAJob.

Depends-On: https://github.com/openshift/release/pull/78116

## Summary by CodeRabbit

* **Tests**
  * Expanded DRA end-to-end coverage with isolated generated namespaces and richer scenario groups (quota, workload types, misconfigured mappings, preemption, gang scheduling, vendor-neutral driver).
  * Reorganized tests into focused contexts, stabilized client usage, and strengthened lifecycle/teardown checks.
  * Added helpers for creating DRA workloads, asserting admission/claim behavior, waiting for quota release, and configuring DRA resources on ClusterQueues.
  * Added an admission synchronization step for Deployment/StatefulSet-style workloads to improve reliability.

### DIFF
--- a/test/e2e/e2e_dra_test.go
+++ b/test/e2e/e2e_dra_test.go
@@ -41,8 +41,7 @@ const (
 	draDeviceClassName     = "gpu.nvidia.com"
 	draLogicalResource     = "nvidia-gpu"
 	draTestNamespacePrefix = "kueue-dra-test-"
-	draTestQueue           = "dra-test-queue"
-	draClusterQueueName    = "dra-test-clusterqueue"
+	draLocalQueueName      = "dra-test-queue"
 )
 
 // setDRAJobCPU reduces CPU request for DRA test jobs to fit on GPU nodes
@@ -51,15 +50,10 @@ func setDRAJobCPU(job *batchv1.Job) {
 	job.Spec.Template.Spec.Containers[0].Resources.Requests[corev1.ResourceCPU] = resource.MustParse("100m")
 }
 
-var _ = Describe("DRA Structured Parameters", Label("operator", "dra"), Ordered, func() {
+var _ = Describe("DRA Structured Parameters", Label("operator", "dra_alice"), Ordered, func() {
 	var (
-		draSupported          bool
-		draTestNamespace      string
-		initialKueueInstance  *ssv1.Kueue
-		kueueClient           *upstreamkueueclient.Clientset
-		cleanupClusterQueue   func()
-		cleanupLocalQueue     func()
-		cleanupResourceFlavor func()
+		draSupported         bool
+		initialKueueInstance *ssv1.Kueue
 	)
 
 	JustAfterEach(func(ctx context.Context) {
@@ -134,47 +128,9 @@ var _ = Describe("DRA Structured Parameters", Label("operator", "dra"), Ordered,
 			return nil
 		}, testutils.OperatorReadyTime, testutils.OperatorPoll).Should(Succeed())
 
-		ns, err := kubeClient.CoreV1().Namespaces().Create(ctx, &corev1.Namespace{
-			ObjectMeta: metav1.ObjectMeta{
-				GenerateName: draTestNamespacePrefix,
-				Labels: map[string]string{
-					testutils.OpenShiftManagedLabel: "true",
-				},
-			},
-		}, metav1.CreateOptions{})
-		Expect(err).NotTo(HaveOccurred())
-		draTestNamespace = ns.Name
-
-		kueueClient = clients.UpstreamKueueClient
-		cleanupResourceFlavor, err = testutils.NewResourceFlavor().Create(ctx, kueueClient)
-		Expect(err).NotTo(HaveOccurred())
-
-		cq := testutils.NewClusterQueue()
-		cq.Name = draClusterQueueName
-		cq.WithDRAResource(draLogicalResource, "1")
-		cleanupClusterQueue, err = cq.Create(ctx, kueueClient)
-		Expect(err).NotTo(HaveOccurred())
-
-		lq := testutils.NewLocalQueue(draTestNamespace, draTestQueue)
-		lq.Spec.ClusterQueue = kueuev1beta2.ClusterQueueReference(draClusterQueueName)
-		cleanupLocalQueue, err = lq.Create(ctx, kueueClient)
-		Expect(err).NotTo(HaveOccurred())
 	})
 
 	AfterAll(func(ctx context.Context) {
-		if cleanupLocalQueue != nil {
-			cleanupLocalQueue()
-		}
-		if cleanupClusterQueue != nil {
-			cleanupClusterQueue()
-		}
-		if cleanupResourceFlavor != nil {
-			cleanupResourceFlavor()
-		}
-
-		err := kubeClient.CoreV1().Namespaces().Delete(ctx, draTestNamespace, metav1.DeleteOptions{})
-		Expect(err).NotTo(HaveOccurred())
-
 		if initialKueueInstance != nil {
 			applyKueueConfig(ctx, initialKueueInstance.Spec.Config, kubeClient)
 		}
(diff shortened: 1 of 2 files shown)


anahas-redhat:
@PannagaRao can you ptal? Change for creating ResourceFlavor, ClusterQueue, Namespace and LocalQueue in each It is implemented.

test/e2e/e2e_dra_test.go
```diff
@@ -129,367 +128,832 @@ var _ = Describe("DRA Structured Parameters", Label("operator", "dra"), Ordered,
...
+				assignment := wl.Status.Admission.PodSetAssignments[0]
+				g.Expect(assignment.ResourceUsage).To(HaveKey(corev1.ResourceName(draLogicalResource)))
+				g.Expect(assignment.ResourceUsage[corev1.ResourceName(draLogicalResource)]).To(Equal(resource.MustParse("1")))
+			}, testutils.OperatorReadyTime, testutils.OperatorPoll).Should(Succeed())
+
+			By("Verifying job is unsuspended")
+			Eventually(func(g Gomega) {
+				j, err := kubeClient.BatchV1().Jobs(ns.Name).Get(ctx, createdJob.Name, metav1.GetOptions{})
+				g.Expect(err).NotTo(HaveOccurred())
+				g.Expect(j.Spec.Suspend == nil || !*j.Spec.Suspend).To(BeTrue())
```
PannagaRao: Could you refactor all tests to use IsJobSuspended() instead of verifying from spec alone?

test/e2e/e2e_dra_test.go
```diff
@@ -129,367 +128,832 @@ var _ = Describe("DRA Structured Parameters", Label("operator", "dra"), Ordered,
...
+			checkWorkloadCondition(ctx, ns.Name, string(createdLowJob.UID), kueuev1beta2.WorkloadAdmitted, "low-priority")
+		})
+	})
 
-		By("Verifying second job is suspended (quota full)")
-		Consistently(func() bool {
-			return testutils.IsJobSuspended(ctx, kubeClient, draTestNamespace, createdJob2.Name)
-		}, testutils.ConsistentlyTimeout, testutils.ConsistentlyPoll).Should(BeTrue())
+	When("gang scheduling is enabled", func() {
+		It("should enforce all-or-nothing admission for gang DRA workloads", func(ctx context.Context) {
```
PannagaRao, line 865: Can we add a skip condition if the number of GPU's present < 2 ?

PannagaRao (changes requested):

test/e2e/e2e_dra_test.go
```diff
@@ -48,13 +50,10 @@ func setDRAJobCPU(job *batchv1.Job) {
 	job.Spec.Template.Spec.Containers[0].Resources.Requests[corev1.ResourceCPU] = resource.MustParse("100m")
 }
 
-var _ = Describe("DRA Structured Parameters", Label("operator", "dra"), Ordered, func() {
+var _ = Describe("DRA Structured Parameters", Label("operator", "dra_alice"), Ordered, func() {
```
PannagaRao: Use the label dra to make sure the CI doesn't skip the tests

PannagaRao (changes requested):

PannagaRao:
All DRA tests were skipped in `test-e2e-dra-gpu-4-21` due to the incorrect label (dra-alice) that has been added. @anahas-redhat you might need to run it again after correcting the label.

test/e2e/e2e_dra_test.go
```diff
@@ -129,367 +128,842 @@ var _ = Describe("DRA Structured Parameters", Label("operator", "dra"), Ordered,
...
-						g.Expect(res.Total.Cmp(resource.MustParse("0"))).To(Equal(0),
-							"%s quota should be 0 before test starts, got %s", draLogicalResource, res.Total.String())
-					}
+	When("gang scheduling is enabled", func() {
+		It("should enforce all-or-nothing admission for gang DRA workloads", func(ctx context.Context) {
+			slices, err := kubeClient.ResourceV1().ResourceSlices().List(ctx, metav1.ListOptions{})
+			Expect(err).NotTo(HaveOccurred())
+			gpuCount := 0
+			for _, s := range slices.Items {
+				if s.Spec.Driver == draDeviceClassName {
```
PannagaRao: This logic will produce incorrect results when there are MIG resource slices in the cluster. You will need to filter based on type. Please refer to the extended resources test.

PannagaRao (changes requested):

test/e2e/e2e_dra_test.go
anahas-redhat: @PannagaRao, sorry, I did not follow. I'm not configuring MIG or even testing it for now. What do you mean? 
Do you see a problem in the future?

PannagaRao: Yes, this filtering might not work if the cluster has MIG profiles as they would also have the same driver. So filtering it based on just the driver name will fetch wrong gpu values

anahas-redhat:
@PannagaRao, can you ptal?

test/e2e/testutils/utils.go
```diff
@@ -846,16 +857,11 @@ func IsJobPodRunning(ctx context.Context, kubeClient *kubernetes.Clientset, name
 }
 
 func IsJobSuspended(ctx context.Context, kubeClient *kubernetes.Clientset, namespace, jobName string) bool {
-	job, err := kubeClient.BatchV1().Jobs(namespace).Get(context.TODO(), jobName, metav1.GetOptions{})
```
PannagaRao, line 849: I actually intended the refactor to use IsJobSuspended() so that the status condition is checked instead of just spec. I don't think we need to change what the function does as checking only the spec can cause flakes.

PannagaRao (changes requested):

test/e2e/testutils/utils.go
anahas-redhat: Actually, checking status.conditions was causing the test to flake. The Suspended status condition is set asynchronously by the Job controller — it's not present immediately after job creation. When Consistently polls at 1 second and the condition hasn't been written yet, the function returns false and the assertion fails instantly.

spec.suspend is the source of truth that Kueue directly controls. It's set at creation time and is immediately available — no async dependency on the Job controller's reconciliation. It can't flake because the value is present from the moment the API server responds to Create().

The CI proved this the test failed 3/3 attempts with the status check, and passed immediately after switching to spec: [PR #1795 job failure](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_kueue-operator/1795/pull-ci-openshift-kueue-operator-main-test-e2e-dra-gpu-4-21/).

If you have a different approach that handles the async timing reliably, I'm open to it.

PannagaRao: Fair point, I understand the flake with Consistently — the condition isn't set immediately after creation. But this function is used across multiple test files for both Consistently and Eventually patterns. I'm concerned that changing it to Spec.Suspend for everyone could introduce the race Kevin flagged for Eventually callers - https://github.com/openshift/kueue-operator/pull/1803#discussion_r3119365220

anahas-redhat: I see... the concern that the Job controller hasn't reacted to the spec change doesn't cause a flake in IsJobSuspended itself — it's about what happens after. In our tests, Eventually(!IsJobSuspended(...)) is always immediately followed by Eventually(IsJobPodRunning(...)), which retries until pods are actually running. So that gap is already handled by the next assertion.

What IsJobSuspended answers is "has Kueue made its admission decision?" — and spec.suspend is the direct signal for that. If we need to also verify the Job controller acknowledged it, that's a separate check (and we already have it via IsJobPodRunning).

PannagaRao: Makes sense — if all the Eventually unsuspend checks are followed by IsJobPodRunning as you mentioned, the gap is already covered.
LGTM from my side, but I'll defer to @kannon92  in case the upstream flakes he mentioned were about a different scenario.

PannagaRao: /lgtm
/approve

/hold

PannagaRao (approved):

test/e2e/testutils/utils.go
kannon92: https://github.com/kubernetes/kubernetes/blob/master/pkg/registry/batch/job/strategy.go#L194

For the job controller we do consider both suspend = true and condition as a gate for doing any mutations.

I'd prefer to keep this function the same.

kannon92:
## Review: `e2e_dra_test.go` — Maintainability & Flakiness

> **Note:** This review was AI-generated by Claude.

### Flakiness Issues

**1. `Consistently` timeout is dangerously short (HIGH)**

`ConsistentlyTimeout` is only **5 seconds** with a 1-second poll interval. This means the "should remain suspended" assertions at lines 278, 342, 770, and 943 only sample ~5 times. A brief delay in Kueue's reconciliation loop could cause the job to still appear suspended when it's actually about to be admitted. If Kueue processes the workload at second 6, the test passes when it should have failed.

**Recommendation:** Increase to at least 15–30 seconds for `Consistently` checks that verify "this should NOT happen." The current 5s is too tight for a controller with a 10s poll interval.

---

**2. Preemption re-admission is timing-sensitive (MEDIUM)**

Lines 856–859: After the high-priority job finishes, the test immediately checks that the low-priority job is re-admitted. The sequence of `WorkloadFinished` (high-prio) → `WorkloadAdmitted` (low-prio) has no explicit ordering guarantee. If the low-priority workload was already re-admitted *before* the high-priority one finished (due to a race), the assertion passes for the wrong reason.

**Recommendation:** Consider verifying that the low-priority workload was evicted *and then* re-admitted (checking `Admitted` became False, then True again) rather than just checking it's True at the end.

---

**3. `sleep 30` in job commands creates a 30-second floor (LOW)**

`newDRAJob` at line 1063 hardcodes `sleep 30`. Tests that wait for `WorkloadFinished` (e.g., gang scheduling line 959) must wait at least 30 seconds for the job to complete. This adds unnecessary wall-clock time.

**Recommendation:** Consider `sleep 5` or `sleep 10` — long enough for assertions to run, short enough to not inflate CI time.

---

**4. Job1 cleanup gap in "admit waiting job" test (MEDIUM)**

Lines 319–347: `createdJob1` has no `defer` cleanup. It's explicitly deleted at line 347 as part of the test logic. If any assertion between lines 324–344 fails, `createdJob1` leaks (the job stays running, holding GPU quota). This could cause subsequent tests to fail due to exhausted quota.

**Recommendation:** Add a `defer testutils.CleanUpJob(...)` immediately after creation as a safety net.

---

### Maintainability Issues

**5. Massive per-It setup duplication (HIGH)**

Every single `It` block repeats the same ~25-line sequence creating ResourceFlavor, ClusterQueue, Namespace, LocalQueue, and ResourceClaimTemplate. This appears **9 times**.

**Recommendation:** A helper like `setupDRATestResources(ctx, kueueClient, kubeClient, draLogicalResource, quota) (ns, lqName, cleanup)` would dramatically reduce noise and make the actual test logic visible.

---

**6. Inconsistent cleanup patterns (MEDIUM)**

Some resources use `DeferCleanup(cleanupFn)` while others use `defer testutils.CleanUpJob(...)`. Within the same test, both patterns are sometimes used.

**Recommendation:** Pick one convention. `DeferCleanup` is preferred in Ginkgo v2 since it integrates with the test lifecycle and runs even on panic.

---

**7. ResourceClaimTemplate never explicitly cleaned up (LOW)**

`ResourceClaimTemplates` are created in each test but never have explicit cleanup — they rely on namespace deletion to cascade-delete them. Since every other resource gets explicit cleanup, this inconsistency is worth noting.

---

**8. `checkWorkloadCondition` accepts raw `string` for condition type (LOW)**

The function takes `conditionType string`, which means a typo like `"Admited"` would compile and silently never match.

**Recommendation:** Consider accepting a typed constant rather than a bare string.

---

**9. `verifyWorkloadCreated` may race with fast pod completion (LOW)**

At line 567, the Pod test calls `verifyWorkloadCreated` which returns successfully if the workload is either `Admitted` *or* `Finished`. If the pod runs and finishes instantly, the subsequent "Verifying Pod is running" check (line 571) would fail because the pod already completed. The `sleep 30` prevents this for jobs, but worth verifying the Pod's command also sleeps long enough.

---

### Summary

| Priority | Issue | Type |
|----------|-------|------|
| HIGH | `Consistently` 5s timeout too short to catch admission races | Flakiness |
| HIGH | 25-line setup block duplicated 9 times | Maintainability |
| MEDIUM | `createdJob1` has no defer cleanup safety net | Flakiness |
| MEDIUM | Preemption re-admission check doesn't verify evict-then-readmit | Flakiness |
| MEDIUM | Inconsistent `DeferCleanup` vs `defer` patterns | Maintainability |
| LOW | `sleep 30` inflates CI wall time | Flakiness/Perf |
| LOW | ResourceClaimTemplate relies on implicit cascade cleanup | Maintainability |
| LOW | `checkWorkloadCondition` uses untyped string | Maintainability |
| LOW | `verifyWorkloadCreated` may race with fast pod completion | Flakiness |

The two highest-impact changes would be: (1) extracting the per-`It` setup into a shared helper, and (2) increasing `ConsistentlyTimeout` for these DRA tests (or using a test-local value rather than the global 5s constant).

anahas-redhat:
@kannon92 can you ptal?

PR #1803, opened 2026-04-21
Add extended resources support

PannagaRao:

PR Summary

- Enable kueue DRAExtendedResources feature gate when the K8s DRAExtendedResource gate is detected on the cluster. Since DRAExtendedResource is not in openshift/api, it can only be set via CustomNoUpgrade — the operator checks spec.customNoUpgrade.enabled on the FeatureGate CR to determine this.
- Add featuregates to the operator's ClusterRole RBAC and regenerate the bundle CSV to allow reading the FeatureGate CR.                                                                                          
- Decouple DRA API availability check from deviceClassMappings so extended resources work without requiring explicit device class configuration.                                                                  
-  Add e2e_dra_extended_resources_test.go with tests for ER admission, quota enforcement, quota release, ClusterQueue usage, and mixed ER+RCT convergence (the mixed test requires 2+ GPUs on a single node and is skipped otherwise).

## Summary by CodeRabbit

* **New Features**
  * Added support for the DRA Extended Resources feature gate so extended-resource scheduling and accounting can be enabled via cluster feature configuration.

* **Chores**
  * Updated operator permissions to allow reading feature-gate resources for correct feature detection and config application.

* **Tests**
  * Added comprehensive end-to-end tests validating extended-resource admission, reservation accounting, mixed ER+RCT scenarios, and controller config changes.

### DIFF
--- a/pkg/configmap/configmap.go
+++ b/pkg/configmap/configmap.go
@@ -195,7 +195,7 @@ func buildFeatureGates(resources kueue.Resources, frameworks []kueue.KueueIntegr
 	// are available on the cluster. On clusters without DRA support (OCP < 4.21),
 	// the feature gate is not enabled but the deviceClassMappings config is preserved
 	// so it takes effect automatically after a cluster upgrade.
-	if len(resources.DeviceClassMappings) > 0 && draSupported {
+	if (len(resources.DeviceClassMappings) > 0 && draSupported) || draExtendedResourceEnabled {
 		featureGates["DynamicResourceAllocation"] = true
 	}
 
--- a/pkg/configmap/configmap.go
+++ b/pkg/configmap/configmap.go
@@ -32,8 +32,8 @@ import (
 	kueue "github.com/openshift/kueue-operator/pkg/apis/kueueoperator/v1"
 )
 
-func BuildConfigMap(namespace string, kueueCfg kueue.KueueConfiguration, gvrToKind map[string]string, draSupported bool, tlsOpts *configapi.TLSOptions) (*corev1.ConfigMap, error) {
-	config, err := defaultKueueConfigurationTemplate(namespace, kueueCfg, gvrToKind, draSupported, tlsOpts)
+func BuildConfigMap(namespace string, kueueCfg kueue.KueueConfiguration, gvrToKind map[string]string, draSupported bool, draExtendedResourceEnabled bool, tlsOpts *configapi.TLSOptions) (*corev1.ConfigMap, error) {
+	config, err := defaultKueueConfigurationTemplate(namespace, kueueCfg, gvrToKind, draSupported, draExtendedResourceEnabled, tlsOpts)
 	if err != nil {
 		return nil, err
 	}
@@ -187,7 +187,7 @@ func buildResources(resources kueue.Resources) *configapi.Resources {
 	}
 }
 
-func buildFeatureGates(resources kueue.Resources, frameworks []kueue.KueueIntegration, draSupported bool, integrationExtFrameworks []kueue.ExternalFramework, multiKueue *kueue.MultiKueue) map[string]bool {
+func buildFeatureGates(resources kueue.Resources, frameworks []kueue.KueueIntegration, draSupported bool, draExtendedResourceEnabled bool, integrationExtFrameworks []kueue.ExternalFramework, multiKueue *kueue.MultiKueue) map[string]bool {
 	featureGates := map[string]bool{}
 
 	// DynamicResourceAllocation is Alpha in Kueue, so we explicitly enable it
@@ -197,6 +197,14 @@ func buildFeatureGates(resources kueue.Resources, frameworks []kueue.KueueIntegr
 	// so it takes effect automatically after a cluster upgrade.
 	if len(resources.DeviceClassMappings) > 0 && draSupported {
 		featureGates["DynamicResourceAllocation"] = true
+	}
+
+	// DRAExtendedResources is Alpha in Kueue. Enable it when the K8s
+	// DRAExtendedResource feature gate is enabled on the cluster. This allows
+	// workloads to request DRA devices using standard extended resource syntax
+	// (e.g., nvidia.com/gpu: 1) when a DeviceClass has extendedResourceName set.
+	if draExtendedResourceEnabled {
+		featureGates["DRAExtendedResources"] = true
 	}
 
 	// SparkApplicationIntegration is Alpha in Kueue, so we explicitly enable it
@@ -249,7 +257,7 @@ func buildAdmissionFairSharing(admissionFairSharing kueue.AdmissionFairSharing) 
 	return result, nil
 }
 
-func defaultKueueConfigurationTemplate(namespace string, kueueCfg kueue.KueueConfiguration, gvrToKind map[string]string, draSupported bool, tlsOpts *configapi.TLSOptions) (*configapi.Configuration, error) {
+func defaultKueueConfigurationTemplate(namespace string, kueueCfg kueue.KueueConfiguration, gvrToKind map[string]string, draSupported bool, draExtendedResourceEnabled bool, tlsOpts *configapi.TLSOptions) (*configapi.Configuration, error) {
 	admissionFairSharing, err := buildAdmissionFairSharing(kueueCfg.AdmissionFairSharing)
 	if err != nil {
 		return nil, fmt.Errorf("failed to build admission fair sharing: %w", err)
@@ -305,7 +313,7 @@ func defaultKueueConfigurationTemplate(namespace string, kueueCfg kueue.KueueCon
 		WaitForPodsReady:           buildWaitForPodsReady(kueueCfg.GangScheduling),
 		FairSharing:                buildFairSharing(kueueCfg.Preemption),
 		Resources:                  buildResources(kueueCfg.Resources),
-		FeatureGates:               buildFeatureGates(kueueCfg.Resources, kueueCfg.Integrations.Frameworks, draSupported, kueueCfg.Integrations.ExternalFrameworks, kueueCfg.MultiKueue),
+		FeatureGates:               buildFeatureGates(kueueCfg.Resources, kueueCfg.Integrations.Frameworks, draSupported, draExtendedResourceEnabled, kueueCfg.Integrations.ExternalFrameworks, kueueCfg.MultiKueue),
 		MultiKueue:                 mapOperatorMultiKueueToKueue(kueueCfg.MultiKueue, gvrToKind),
 		AdmissionFairSharing:       admissionFairSharing,
 	}, nil
--- a/pkg/configmap/configmap_test.go
+++ b/pkg/configmap/configmap_test.go
@@ -451,7 +451,6 @@ webhook:
 			wantErr: nil,
 		},
 		"dra extended resources enabled": {
-			draSupported:               true,
 			draExtendedResourceEnabled: true,
 			configuration: kueue.KueueConfiguration{
 				Integrations: kueue.Integrations{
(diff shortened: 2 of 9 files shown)


PannagaRao:
Currently, DRAExtendedResources is enabled alongside DynamicResourceAllocation whenever deviceClassMappings are configured and DRA APIs are available. There's no K8s API to query whether a specific feature gate like DRAExtendedResource is enabled on the cluster — the standard pattern across the ecosystem is to check API availability via discovery.
                                                                                                                                                                                                                    
One alternative I considered was listing DeviceClasses during each reconciliation to check if any have extendedResourceName set, and only enabling the gate when one is found. However, this adds API overhead on every reconcile for a gate that is a no-op when no matching DeviceClass exists — kueue's own code path already discovers DeviceClasses with extendedResourceName via a field indexer at runtime and simply treats unmatched extended resources as non-DRA resources.

@kannon92 PTAL and let me know your thoughts!

kannon92:
Is there a way to confirm this tests run on this PR?

/test

PannagaRao:
> Is there a way to confirm this tests run on this PR?
> 
> /test

Nope, there are only periodics that run dra e2es and the CI doesn't patch DRAExtendedResources yet.

kannon92:
> Nope, there are only periodics that run dra e2es and the CI doesn't patch DRAExtendedResources yet.

So I'm worried that we will merge this and then our CI starts failing. Any manual runs of these tests you can perform to give confidence we aren't merging bad tests?

PannagaRao:
> > Nope, there are only periodics that run dra e2es and the CI doesn't patch DRAExtendedResources yet.
> 
> So I'm worried that we will merge this and then our CI starts failing. Any manual runs of these tests you can perform to give confidence we aren't merging bad tests?

I have verified them manually with 4.21 cluster. It had 1 GPU on a worker node so the last test of this PR which expects 2 GPU's at least got skipped as expected. 

```
[SKIPPED] [8.549 seconds]
DRA when Extended Resources are backed by DRA [It] should account both extended resource and ResourceClaimTemplate under same quota [operator, dra, dra-extended-resources]
/home/pbhojara/Documents/git-repo/kueue-operator/test/e2e/e2e_dra_test.go:778

  [SKIPPED] Need at least 2 GPUs on a single node for mixed ER + RCT test
  In [It] at: /home/pbhojara/Documents/git-repo/kueue-operator/test/e2e/e2e_dra_test.go:780 @ 04/20/26 23:38:57.094
------------------------------
SSSSSSSSSSSSSSSSSSSSSSSSSSSSSSS

Ran 8 of 73 Specs in 332.723 seconds
SUCCESS! -- 8 Passed | 0 Failed | 0 Pending | 65 Skipped
```

test/e2e/e2e_dra_test.go
```diff
@@ -77,43 +78,51 @@ var _ = Describe("DRA Structured Parameters", Label("operator", "dra"), Ordered,
...
 		}
 
+		for _, count := range gpusPerNode {
+			if count > maxGPUsPerNode {
+				maxGPUsPerNode = count
+			}
+		}
+		if maxGPUsPerNode == 0 {
+			Skip("No GPU devices found in ResourceSlices, skipping test")
+		}
```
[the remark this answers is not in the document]
kannon92: This is a valid concern albeit quite minor.

WDYT of this suggestion?

We are depending on the NVIDA driver never changing its attributes so leaving some breadcrumbs may be helpful for future us.

kannon92:
good

/hold cancel

test/e2e/e2e_dra_test.go
```diff
@@ -183,313 +192,694 @@ var _ = Describe("DRA Structured Parameters", Label("operator", "dra"), Ordered,
...
+				wl := wlList.Items[0]
+				g.Expect(wl.Status.Admission).NotTo(BeNil(), "workload should be admitted")
+				g.Expect(wl.Status.Admission.PodSetAssignments).To(HaveLen(1))
+
+				assignment := wl.Status.Admission.PodSetAssignments[0]
+				g.Expect(assignment.ResourceUsage).To(HaveKey(corev1.ResourceName(draLogicalResource)))
+				g.Expect(assignment.ResourceUsage[corev1.ResourceName(draLogicalResource)]).To(Equal(resource.MustParse("1")))
+			}, testutils.OperatorReadyTime, testutils.OperatorPoll).Should(Succeed())
+
+			By("Verifying job is unsuspended")
```
kannon92: I think this can be flaky. When checking if a job is unsuspended you should check condition also.

We have had flakes in upstream where the spec can change but the controller won't react to suspend change yet.

PannagaRao:
I have created https://github.com/openshift/release/pull/78132 to add the feature gate. We can verify once this PR merges

test/e2e/e2e_dra_test.go
PannagaRao: Good catch — will replace the Spec.Suspend checks with `testutils.IsJobSuspended()` which verifies the Suspended condition in Status.Conditions rather than just the spec field.

PannagaRao:
As per the discussions, the PR now does the following.

- Enable kueue DRAExtendedResources feature gate when the K8s DRAExtendedResource gate is detected on the cluster. Since DRAExtendedResource is not in openshift/api, it can only be set via CustomNoUpgrade — the operator checks spec.customNoUpgrade.enabled on the FeatureGate CR to determine this.
- Add featuregates to the operator's ClusterRole RBAC and regenerate the bundle CSV to allow reading the FeatureGate CR.                                                                                          
- Decouple DRA API availability check from deviceClassMappings so extended resources work without requiring explicit device class configuration.                                                                  
- Add e2e_dra_extended_resources_test.go with tests for ER admission, quota enforcement, quota release, ClusterQueue usage, and mixed ER+RCT convergence (the mixed test requires 2+ GPUs on a single node and is 
  skipped otherwise).

PannagaRao:
@coderabbitai review

pkg/operator/target_config_reconciler.go
```diff
@@ -338,6 +346,29 @@ func (c *TargetConfigReconciler) sync(ctx context.Context, syncCtx factory.SyncC
...
+	// Unlike DynamicResourceAllocation, this does not require deviceClassMappings.
+	if c.isOpenShift && draAPIsAvailable {
+		fg, err := c.openshiftConfigClient.ConfigV1().FeatureGates().Get(ctx, "cluster", metav1.GetOptions{})
+		if err != nil {
+			klog.Warningf("unable to read FeatureGate CR, preserving previous state: %v", err)
+		} else {
+			c.draExtendedResourceEnabled = false
+			if fg.Spec.FeatureSet == configv1.CustomNoUpgrade && fg.Spec.CustomNoUpgrade != nil {
+				for _, gate := range fg.Spec.CustomNoUpgrade.Enabled {
+					// K8s uses singular "DRAExtendedResource", kueue uses plural "DRAExtendedResources"
```
kannon92, line 362: Oh man this is not confusing at all.

kannon92: @PannagaRao maybe we should change this in upstream?

pkg/configmap/configmap.go
```diff
@@ -199,6 +199,14 @@ func buildFeatureGates(resources kueue.Resources, frameworks []kueue.KueueIntegr
 		featureGates["DynamicResourceAllocation"] = true
 	}
 
+	// DRAExtendedResources is Alpha in Kueue. Enable it when the K8s
+	// DRAExtendedResource feature gate is enabled on the cluster. This allows
+	// workloads to request DRA devices using standard extended resource syntax
+	// (e.g., nvidia.com/gpu: 1) when a DeviceClass has extendedResourceName set.
+	if draExtendedResourceEnabled {
```
kannon92, line 206:
```suggestion
	if draSupported && draExtendedResourceEnabled {
```

pkg/configmap/configmap_test.go
```diff
@@ -442,6 +443,61 @@ resources:
...
     - gpu-large.example.com
     name: example.com/gpus
+webhook:
+  port: 9443
+`,
+				},
+			},
+			wantErr: nil,
+		},
+		"dra extended resources enabled": {
```
kannon92, line 453: You should add a test case for draSupported: False

pkg/configmap/configmap.go
PannagaRao: We don't have to add this check since ExtendedResources do not need deviceClassMappings. We are already checking whether DRA APIs are available before setting draExtendedResourceEnabled.

pkg/configmap/configmap_test.go
PannagaRao: The behavior of ExtendedResources isn't expected to change with value of draSupported. draSupported checks for DRA API being available when DeviceClassMappings are set in the CR whereas draExtendedResourceEnabled is true without DeviceClassMappings when the API's are available and k8s feature gate is enabled. 

I will remove the draSupported value in the test to clear the confusion.

pkg/configmap/configmap.go
kannon92: Hmm okay. The code has a bug though as you can call this function with 1 true and other false.

pkg/configmap/configmap_test.go
kannon92: But this would disable the DRAFeatureGate for Kueue which would be an issue right?

Wouldn't we want both feature gates set in KueueConfig if DRA is true?

test/e2e/e2e_dra_extended_resources_test.go
```diff
@@ -0,0 +1,577 @@
...
+					draSupported = true
+					break
+				}
+			}
+		}
+		if !draSupported {
+			Skip("DRA APIs (resource.k8s.io/v1) not available on this cluster")
+		}
+
+		// Check if ResourceSlices exist for gpu.nvidia.com (driver is running)
```
kannon92: I think you need to put this is an eventually.

We don't really control when this driver gets deployed and goes ready.

test/e2e/e2e_dra_extended_resources_test.go
```diff
@@ -0,0 +1,577 @@
...
+	corev1 "k8s.io/api/core/v1"
+	resourcev1 "k8s.io/api/resource/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/utils/ptr"
+	kueuev1beta2 "sigs.k8s.io/kueue/apis/kueue/v1beta2"
+)
+
+const (
+	erTestNamespace         = "kueue-dra-er-test"
```
kannon92: We should use a new namespace for each test.

You should follow our existing test patterns.

pkg/configmap/configmap.go
PannagaRao: I see your point . I'll update buildFeatureGates to enable DynamicResourceAllocation when either draSupported (mappings + APIs) or draExtendedResourceEnabled (APIs + K8s gate) is true.

@kannon92 WDYT?

pkg/configmap/configmap_test.go
PannagaRao: We would want both feature gates to be true if draExtendedResourceEnabled is enabled actually. I see the bug you pointed out.

test/e2e/e2e_dra_extended_resources_test.go
PannagaRao: Changed it to use eventually

test/e2e/e2e_dra_extended_resources_test.go
PannagaRao: I have updated the tests

pkg/configmap/configmap.go
PannagaRao: I believe to have addressed all the review comments. Could you please take a look @kannon92

kannon92 (commented):
/lgtm

PR #1842, opened 2026-05-04
Fix ExtendedResource test for non-GPU clusters

PannagaRao:

## Summary by CodeRabbit

* **Tests**
  * Updated DRA Extended Resources test initialization to perform a single immediate check for ResourceSlices instead of polling for availability. Tests will now skip immediately if required resources are not found, rather than waiting for driver deployment.


PannagaRao:
cc: @kannon92

PR #1874, opened 2026-05-11
OCPKUEUE-620: Add DRA Extended Resources e2e tests

anahas-redhat:
Implement five test scenarios from the DRA Extended Resources manual test plan, covering Kueue's handling of workloads that use the DRAExtendedResource feature (auto-generated ResourceClaims via DeviceClass.spec.extendedResourceName):

- S1: Workload Types — verifies Pod, Deployment, and StatefulSet workloads are admitted with ER, auto-generated ResourceClaims have the correct ER annotation and DeviceClass, and ClusterQueue tracks nvidia.com/gpu usage.

- S2: Preemption — verifies a high-priority ER Job evicts a low-priority ER Job, ResourceClaims are cleaned up on eviction, and the low-priority Job is re-admitted after the high-priority one completes.

- S3: Gang Scheduling — verifies all-or-nothing admission for multi-pod ER Jobs: a 2-GPU gang Job stays suspended while only 1 GPU is free, then is admitted once the blocking Job is deleted.

- S4: DeviceClass CEL Selector Validation — Part A patches the DeviceClass CEL to match no devices (type=='tpu'), confirms the pod stays Pending while Kueue still admits the workload, then restores the CEL and verifies recovery. Part B creates a custom DeviceClass targeting a specific GPU by UUID and verifies the scheduler allocates exactly that device.

- S5: Remove extendedResourceName — removes the ER mapping from the DeviceClass, confirms no ResourceClaim is auto-generated and the pod cannot schedule, then restores the mapping and verifies recovery with a fresh pod.

## Summary by CodeRabbit

* **Tests**
  * Expanded E2E coverage for DRA extended resources across Pod/Deployment/StatefulSet workloads: automatic resource-claim creation and device-class allocation, preemption and re-admission, gang scheduling (all-or-nothing), selector-based device targeting (including specific GPU), simulated selector mismatch then restore, scenarios with extended-resource declarations removed, tightened suspension assertions, centralized ClusterQueue reservation checks, consistent quota setup and claim naming.

### DIFF
--- a/test/e2e/e2e_dra_extended_resources_test.go
+++ b/test/e2e/e2e_dra_extended_resources_test.go
@@ -31,6 +31,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/utils/ptr"
 	kueuev1beta2 "sigs.k8s.io/kueue/apis/kueue/v1beta2"
+	upstreamkueueclient "sigs.k8s.io/kueue/client-go/clientset/versioned"
 )
 
 const (
@@ -44,6 +45,7 @@ var _ = Describe("DRA Extended Resources", Label("operator", "dra", "dra-extende
 	var (
 		initialKueueInstance *ssv1.Kueue
 		maxGPUsPerNode       int
+		originalDeviceClass  *resourcev1.DeviceClass
 	)
 
 	JustAfterEach(func(ctx context.Context) {
@@ -112,6 +114,7 @@ var _ = Describe("DRA Extended Resources", Label("operator", "dra", "dra-extende
 		if dc.Spec.ExtendedResourceName == nil || *dc.Spec.ExtendedResourceName == "" {
 			Skip("DeviceClass gpu.nvidia.com does not have extendedResourceName set - DRAExtendedResource feature gate not enabled")
 		}
+		originalDeviceClass = dc.DeepCopy()
 
 		kueueInstance, err := clients.KueueClient.KueueV1().Kueues().Get(ctx, "cluster", metav1.GetOptions{})
 		Expect(err).ToNot(HaveOccurred())
@@ -142,8 +145,16 @@ var _ = Describe("DRA Extended Resources", Label("operator", "dra", "dra-extende
 	})
 
 	AfterAll(func(ctx context.Context) {
+		if originalDeviceClass != nil {
+			dc, err := kubeClient.ResourceV1().DeviceClasses().Get(ctx, draDeviceClassName, metav1.GetOptions{})
+			if err == nil {
+				dc.Spec.ExtendedResourceName = originalDeviceClass.Spec.ExtendedResourceName
+				dc.Spec.Selectors = originalDeviceClass.Spec.Selectors
+				_, _ = kubeClient.ResourceV1().DeviceClasses().Update(ctx, dc, metav1.UpdateOptions{})
+			}
+		}
 		if initialKueueInstance != nil {
-			applyKueueConfig(ctx, initialKueueInstance.Spec.Config, kubeClient)
+			restoreKueueConfig(ctx, initialKueueInstance.Spec.Config, kubeClient)
 		}
 	})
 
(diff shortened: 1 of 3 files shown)


anahas-redhat:
Hey @PannagaRao and @sohankunkerkar. Can you PTAL?
Thanks.

test/e2e/e2e_dra_extended_resources_test.go
```diff
@@ -311,9 +304,10 @@ var _ = Describe("DRA Extended Resources", Label("operator", "dra", "dra-extende
 		defer testutils.CleanUpJob(ctx, kubeClient, createdJob.Namespace, createdJob.Name)
 
 		By("Verifying job remains suspended")
-		Consistently(func() bool {
-			return testutils.IsJobSuspended(ctx, kubeClient, ns.Name, createdJob.Name)
-		}, testutils.ConsistentlyTimeout, testutils.ConsistentlyPoll).Should(BeTrue())
+		Eventually(func(g Gomega) {
```
sohankunkerkar, line 308: Any reason to use Eventually here? I'm asking bc it can't detect a job that's briefly suspended and then wrongly unsuspended.

test/e2e/e2e_dra_extended_resources_test.go
```diff
@@ -394,22 +380,25 @@ var _ = Describe("DRA Extended Resources", Label("operator", "dra", "dra-extende
 		defer testutils.CleanUpJob(ctx, kubeClient, createdJob2.Namespace, createdJob2.Name)
 
 		By("Verifying second job is suspended (quota full)")
-		Consistently(func() bool {
-			return testutils.IsJobSuspended(ctx, kubeClient, ns.Name, createdJob2.Name)
-		}, testutils.ConsistentlyTimeout, testutils.ConsistentlyPoll).Should(BeTrue())
+		Eventually(func(g Gomega) {
```
sohankunkerkar, line 390: The same: https://github.com/openshift/kueue-operator/pull/1874/changes#r3315158603

test/e2e/e2e_dra_extended_resources_test.go
```diff
@@ -561,115 +554,1066 @@ var _ = Describe("DRA Extended Resources", Label("operator", "dra", "dra-extende
...
+				for _, claim := range claims.Items {
+					if claim.Status.Allocation != nil && len(claim.Status.ReservedFor) > 0 {
+						g.Expect(claim.Annotations).To(HaveKeyWithValue("resource.kubernetes.io/extended-resource-claim", "true"))
+						g.Expect(claim.Spec.Devices.Requests).NotTo(BeEmpty())
+						g.Expect(claim.Spec.Devices.Requests[0].Exactly).NotTo(BeNil())
+						g.Expect(claim.Spec.Devices.Requests[0].Exactly.DeviceClassName).To(Equal(draDeviceClassName))
+						foundValid = true
+					}
+				}
+				g.Expect(foundValid).To(BeTrue(), "no valid allocated/reserved ResourceClaim found for high-priority job")
```
sohankunkerkar: The low-priority claim check above correctly filters by pod UID via OwnerReferences, but this one matches ANY allocated+reserved claim in the namespace. After eviction the old claim may still exist during  async GC. Use the same OwnerReferences filter pattern here and on the re-admitted check below.

sohankunkerkar:
I think this diff is pretty huge. Consider extracting the repeated setup (ResourceFlavor → CQ → Namespace → LQ) and ResourceClaim verification into shared helpers if possible. This would cut the diff roughly in half and make each test case read as just its scenario-specific logic.

test/e2e/e2e_dra_extended_resources_test.go
anahas-redhat: The reason I changed these from Consistently to Eventually is that Consistently was causing flakes in CI. The likely root cause is that IsJobSuspended can return false transiently (e.g. API timeout or Kueue hasn't reconciled the workload yet), which causes Consistently to fail immediately on the first bad poll.
 
I can suggest maybe a couple different things: 
1. Change from "Verifying job remains suspended" to "Verifying job is suspended".
2. First Eventually to wait for Kueue to reconcile and confirm the job IS suspended (handles the startup race) -> Then Consistently to verify it stays suspended over the window

WDYT? Or, is there any other approach you see here?

test/e2e/e2e_dra_extended_resources_test.go
anahas-redhat: Same fix would be applied here.

test/e2e/e2e_dra_extended_resources_test.go
anahas-redhat: Good catch — I'll update both the high-priority and re-admitted low-priority claim checks to filter by pod UID via OwnerReferences.

anahas-redhat:
> I think this diff is pretty huge. Consider extracting the repeated setup (ResourceFlavor → CQ → Namespace → LQ) and ResourceClaim verification into shared helpers if possible. This would cut the diff roughly in half and make each test case read as just its scenario-specific logic.

Thanks for the suggestion. The repeated setup is intentional to stay consistent with the existing e2e test patterns — every other test file (e2e_dra_test.go, e2e_preemption_test.go, e2e_gangscheduling_test.go, e2e_admission_fair_sharing_test.go, etc.) repeats the ResourceFlavor → ClusterQueue → Namespace → LocalQueue setup inline in each It() block rather than using shared helpers. Happy to refactor, but I think that should be done as a separate cleanup across all e2e files to keep things consistent, rather than just in this PR. WDYT?

kannon92:
> Thanks for the suggestion. The repeated setup is intentional to stay consistent with the existing e2e test patterns — every other test file (e2e_dra_test.go, e2e_preemption_test.go, e2e_gangscheduling_test.go, e2e_admission_fair_sharing_test.go, etc.) repeats the ResourceFlavor → ClusterQueue → Namespace → LocalQueue setup inline in each It() block rather than using shared helpers. Happy to refactor, but I think that should be done as a separate cleanup across all e2e files to keep things consistent, rather than just in this PR. WDYT?

Please do this cleanup. This repeating of code is making reviews of these tests really difficult.

anahas-redhat:
@sohankunkerkar I went ahead and did the fixes. Let me know if you're ok with them:

- Filter ResourceClaim checks by pod UID via OwnerReferences: the high-priority and re-admitted low-priority checks were matching ANY allocated+reserved claim in the namespace, which could match stale claims during async GC after eviction. Use the same OwnerReferences filter pattern (matching by pod UID) that the low-priority claim check already uses.

- Use two-phase Eventually/Consistently for suspension checks: first Eventually to wait for Kueue to reconcile and confirm the job IS suspended (handles the startup race), then Consistently to verify it stays suspended over the window.

- Consider extracting the repeated setup (ResourceFlavor → CQ → Namespace → LQ) and ResourceClaim verification into shared helpers if possible. - it will be address in a new PR.

sohankunkerkar (approved):
Thanks!
/lgtm

PR #1883, opened 2026-05-13
OCPKUEUE-642: DRA (structured param) + Admission Fair Sharing tests

anahas-redhat:
Add two new e2e tests for Admission Fair Sharing (AFS) with DRA:
- Verify DRA GPU usage appears in consumedResources for the GPU-using LocalQueue
- Verify AFS deprioritizes GPU-heavy queue in admission ordering when CPU quota is exhausted

Also fix flakiness in config restoration during test cleanup by introducing restoreKueueConfig, which retries Get+Update on resourceVersion conflicts caused by the operator reconciler modifying the Kueue CR concurrently. All DeferCleanup and AfterAll sites that restore Kueue config now use this helper.

## Summary by CodeRabbit

* **Tests**
  * Added end-to-end tests validating Admission Fair Sharing with DRA/GPU usage and admission ordering between GPU- and CPU-focused queues.
  * New cases confirm GPU logical resources affect GPU queue consumption and not CPU-only queues; GPU-heavy queues can be deprioritized under contention.
  * Added scenarios covering misconfigured device-class mappings and gang scheduling.
  * Standardized test cleanup, added retry/backoff and readiness checks for stable config restores.
  * Switched a container image to the standardized test workload image.

### DIFF
--- a/test/e2e/e2e_dra_test.go
+++ b/test/e2e/e2e_dra_test.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"fmt"
 	"strings"
+	"time"
 
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
@@ -76,14 +77,20 @@ var _ = Describe("DRA Structured Parameters", Label("operator", "dra"), Ordered,
 			Skip("DRA APIs (resource.k8s.io/v1) not available on this cluster")
 		}
 
-		// Check if ResourceSlices exist for gpu.nvidia.com (NVIDIA DRA driver running)
-		slices, err := kubeClient.ResourceV1().ResourceSlices().List(ctx, metav1.ListOptions{})
-		Expect(err).NotTo(HaveOccurred())
+		// Wait for ResourceSlices from gpu.nvidia.com (NVIDIA DRA driver may still be deploying)
+		var slices *resourcev1.ResourceSliceList
 		hasDriverSlices := false
-		for _, s := range slices.Items {
-			if s.Spec.Driver == draDeviceClassName {
-				hasDriverSlices = true
-				break
+		for i := 0; i < 6 && !hasDriverSlices; i++ {
+			if i > 0 {
+				time.Sleep(5 * time.Second)
+			}
+			slices, err = kubeClient.ResourceV1().ResourceSlices().List(ctx, metav1.ListOptions{})
+			Expect(err).NotTo(HaveOccurred())
+			for _, s := range slices.Items {
+				if s.Spec.Driver == draDeviceClassName {
+					hasDriverSlices = true
+					break
+				}
 			}
 		}
 		if !hasDriverSlices {
@@ -152,8 +159,10 @@ var _ = Describe("DRA Structured Parameters", Label("operator", "dra"), Ordered,
 
 	AfterAll(func(ctx context.Context) {
 		if initialKueueInstance != nil {
+			// Clear deviceClassMappings rather than restoring originals to avoid
+			// contaminating the extended-resources suite which manages its own mappings.
 			initialKueueInstance.Spec.Config.Resources.DeviceClassMappings = nil
-			applyKueueConfig(ctx, initialKueueInstance.Spec.Config, kubeClient)
+			restoreKueueConfig(ctx, initialKueueInstance.Spec.Config, kubeClient)
 		}
 	})
 
@@ -199,7 +208,7 @@ var _ = Describe("DRA Structured Parameters", Label("operator", "dra"), Ordered,
 			job := newDRAJob(builder, "dra-basic-job", "gpu-template-basic", draLocalQueueName)
 			createdJob, err := kubeClient.BatchV1().Jobs(ns.Name).Create(ctx, job, metav1.CreateOptions{})
 			Expect(err).NotTo(HaveOccurred())
-			defer testutils.CleanUpJob(ctx, kubeClient, createdJob.Namespace, createdJob.Name)
+			DeferCleanup(testutils.CleanUpJob, kubeClient, createdJob.Namespace, createdJob.Name)
 
 			By("Verifying workload is admitted with correct DRA quota")
 			Eventually(func(g Gomega) {
@@ -270,7 +279,7 @@ var _ = Describe("DRA Structured Parameters", Label("operator", "dra"), Ordered,
 			job := newDRAJob(builder, "dra-exceed-job", "gpu-template-exceed", draLocalQueueName)
 			createdJob, err := kubeClient.BatchV1().Jobs(ns.Name).Create(ctx, job, metav1.CreateOptions{})
 			Expect(err).NotTo(HaveOccurred())
-			defer testutils.CleanUpJob(ctx, kubeClient, createdJob.Namespace, createdJob.Name)
+			DeferCleanup(testutils.CleanUpJob, kubeClient, createdJob.Namespace, createdJob.Name)
 
 			By("Verifying job is suspended due to exceeding DRA quota")
 			Eventually(func() bool {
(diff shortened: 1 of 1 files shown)


anahas-redhat:
Hi @kannon92 @sohankunkerkar when it's possible, can you ptal?
Thanks in advance.

kannon92:
> **Note:** This review was generated with AI assistance (Claude). Findings should be validated by a human reviewer.

---

## Review: DRA AFS e2e tests — Flakiness, Correctness, and DRY

### Flakiness

**1. Half-life decay race in AFS ordering test (HIGH)**

`e2e_dra_test.go:1218-1313` — `UsageHalfLifeTimeSeconds: 30` is configured at the `When` block level, but the ordering test spends significant wall-clock time between building GPU usage history and the final admission ordering assertion:
- Wait for jobs to complete (variable)
- Verify usage persists (up to 30s)
- Create placeholder, wait for admission (up to `OperatorReadyTime`)
- Create contenders, `Consistently` 10s
- Delete placeholder, check ordering

By the time the placeholder is deleted and the contenders compete, the 30-second half-life may have decayed lq-gpu's GPU usage advantage enough that ordering becomes non-deterministic. The final `Consistently` at line 1310 only watches for 15 seconds, which is thin margin.

**Suggestion**: Increase `UsageHalfLifeTimeSeconds` (e.g., 300) so decay is negligible during the test window, or reduce the number of sequential steps between building usage and testing ordering.

**2. ResourceSlice detection changed from Eventually to one-shot (MEDIUM)**

`e2e_dra_test.go:80-92` — The old code used `Eventually` to wait for ResourceSlices, handling the case where the NVIDIA DRA driver is still deploying. The new code does a single `List` and `Skip`s immediately if none are found. In CI environments where the driver starts slowly or pods are still rolling out, this will produce false skips.

**Suggestion**: Keep an `Eventually` with a shorter timeout (e.g., 30s) rather than skipping on first check.

**3. `restoreKueueConfig` doesn't wait for ConfigMap reconciliation (MEDIUM)**

`e2e_dra_test.go:1426-1451` — Unlike `applyKueueConfig` (which waits for the deployment resourceVersion to change), `restoreKueueConfig` only waits for replicas to be ready. If the restore doesn't trigger a visible deployment update, the function may return before the ConfigMap actually reflects the restored values. A subsequent test's `BeforeAll` could then see stale config.

**Suggestion**: Add a resourceVersion-change wait or verify the ConfigMap content matches the restored config, similar to what `applyKueueConfig` does.

---

### Correctness

**4. AfterAll force-clears DeviceClassMappings regardless of original state (LOW)**

`e2e_dra_test.go:156` — `initialKueueInstance.Spec.Config.Resources.DeviceClassMappings = nil` overwrites whatever the original cluster had. `initialKueueInstance` was saved as a DeepCopy of the pre-test config at line 112, so this doesn't truly restore — it restores everything *except* deviceClassMappings. If this is intentional (to avoid contaminating the extended-resources suite), a short comment explaining why would help future readers.

**5. Outer AfterAll uses `applyKueueConfig` without conflict retry (LOW)**

`e2e_dra_test.go:157` — The inner AFS `AfterAll` uses `restoreKueueConfig` (with retry), but the outer suite `AfterAll` still uses `applyKueueConfig` which has no retry on resourceVersion conflicts. In an `Ordered` suite this is likely fine since the outer runs last, but it's inconsistent with the pattern established by `restoreKueueConfig`.

---

### DRY

The two new AFS `It` blocks share large stretches of nearly identical code. Four patterns in particular:

**6. AFS test infrastructure setup (lines 1010-1044 vs 1128-1162)**

The ResourceFlavor + CQ-with-AFS + Namespace + two LocalQueues block is copied verbatim. This could be lifted into a helper or the `When` block's `BeforeEach`:

```go
func setupAFSInfrastructure(ctx context.Context, kueueClient *upstreamkueueclient.Clientset, templateName string) (cq *kueuev1beta2.ClusterQueue, ns *corev1.Namespace, lqGPU, lqCPU *kueuev1beta2.LocalQueue)
```

**7. Submit 2 GPU jobs + 1 CPU job and wait for running (lines 1051-1080 vs 1169-1198)**

The pattern of creating two `newDRAJob`s on lq-gpu, one CPU job on lq-cpu, and the `Eventually`-running loop is copy-pasted with only name suffixes changing. A helper like:

```go
func submitAFSWorkload(ctx context.Context, ns, lqGPU, lqCPU, templateName string) (gpu1, gpu2, cpuJob *batchv1.Job)
```

**8. Verify N allocated ResourceClaims (lines 1082-1093 vs 1200-1211)**

Identical `Eventually` block counting allocated+reserved ResourceClaims. `verifyResourceClaimAllocated` already exists at line 1383 but only checks `claims.Items[0]` — it could be generalized with a count parameter to cover both uses:

```go
func verifyAllocatedResourceClaims(ctx context.Context, namespace string, expectedCount int)
```

**9. CPU-only job construction with explicit resources (lines 1234-1245, 1257-1268, 1273-1284)**

The placeholder, gpu-compete, and cpu-compete jobs all follow the same pattern: set name, queue label, command, and identical CPU/Memory requests+limits. This appears 3 times in a single test. A small helper would clean it up:

```go
func newCPUOnlyJob(builder *testutils.TestResourceBuilder, name, queue, command, cpu, memory string) *batchv1.Job
```

Items 6 and 7 are the biggest wins — together they'd remove ~60 lines of duplication.

---

### Good changes worth calling out

- Switching from `defer testutils.CleanUpJob` to `DeferCleanup` is correct — `defer` in Ginkgo `It` blocks captures the context at the wrong scope and can silently fail.
- The new `restoreKueueConfig` with conflict retry fixes a real flakiness source in cleanup paths.
- The ClusterQueue reservation verification before creating the second job (lines 335-349) eliminates a race where the second job could be created before the first's reservation is visible.
- The cross-suite `deviceClassMappings` cleanup in the extended-resources `BeforeAll` (line 119) is a good fix for inter-suite ordering issues.

anahas-redhat:
Thanks for the review! Here's what I've addressed:

**Flakiness**
1. **Half-life decay race (HIGH)** - Increased `UsageHalfLifeTimeSeconds` from `30` to `120`. At ~90s worst-case elapsed time, this retains ~63% of usage - sufficient margin for deterministic ordering.
2. **ResourceSlice detection one-shot (MEDIUM)** - Changed from single `List` to a polling loop (~30s) to handle slow NVIDIA DRA driver startup in CI.
3. **restoreKueueConfig ConfigMap reconciliation (MEDIUM)** - Added deployment `resourceVersion`-change wait before the readiness check, matching `applyKueueConfig` behavior.

**Correctness**
4. **AfterAll force-clears DeviceClassMappings (LOW)** - Added comment explaining the intentional nil-out (avoids contaminating extended-resources suite).
5. **Outer AfterAll uses applyKueueConfig without retry (LOW)** - Switched to `restoreKueueConfig` for consistency and retry safety.

**CodeRabbit pre-merge checks**
- Replaced bare `"busybox"` image with `testutils.GetContainerImageForWorkloads()` for disconnected environment compatibility.
- Added doc comment to `newDRAResourceClaimTemplate` to bring docstring coverage above 80%.

**DRY (items 6-9)** - Not addressed in this iteration. The inline setup pattern (ResourceFlavor + ClusterQueue + Namespace + LocalQueue in each `It` block) is consistent with the existing convention across the e2e suite — `e2e_gangscheduling_test.go` and `e2e_preemption_test.go` follow the same structure. Extracting helpers would diverge from this established pattern. Happy to refactor if you feel strongly about it.

cc @kannon92

sohankunkerkar (approved):
/lgtm

anahas-redhat:
@sohankunkerkar can you please give it another look after https://github.com/openshift/kueue-operator/pull/1899 changes? 

Just in case, I ran the Kueue-DRA suite again, all tests have [passed](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift_kueue-operator/1883/pull-ci-openshift-kueue-operator-main-test-e2e-dra-gpu-4-21/2056844118453129216/artifacts/test-e2e-dra-gpu-4-21/e2e-kueue-dra/build-log.txt).

sohankunkerkar (approved):
/lgtm

PR #1886, opened 2026-05-14
OCPBUGS-85644: Fix DRA e2e test failures on periodic CI jobs

anahas-redhat:
Since May 7, the periodic-ci-openshift-kueue-operator-main-test-e2e-dra-gpu-4-21 job has been failing on every daily run. Two problems have
  been identified:

**Problem 1** — ER resource naming mismatch (100% failure rate since May 7): The ER test creates ClusterQueues using nvidia.com/gpu, but Kueue translates it to nvidia-gpu because deviceClassMappings leaked from e2e_dra_test.go's BeforeAll, which runs first under the shared dra  label. Workloads can never match flavors.
  - Fix: Clear deviceClassMappings in ER test BeforeAll and wait for config reconciliation before proceeding.

**Problem 2** — DRA quota test race condition (intermittent, May 10-11): The quota enforcement test creates the second job before the ClusterQueue status reflects the first job's reservation.
  - Fix: Add Eventually check that FlavorsReservation shows the first job's reservation before creating the second job.

  Additionally, since May 8, periodic-ci-openshift-kueue-operator-main-test-e2e-downstream-4-22 has failed on every daily run:

 **Problem 3** — DRA Structured Parameters BeforeAll hard-fails without NVIDIA driver (100% failure rate since May 8): Uses Eventually with a 180s timeout waiting for ResourceSlices, failing hard on clusters without the NVIDIA DRA driver. The extended resources test handles this correctly with an immediate check and Skip().
  - Fix: Replace Eventually timeout with immediate check + Skip(), matching the extended resources test pattern. 
  
  Affected Prow Jobs:
  - [periodic-ci-openshift-kueue-operator-main-test-e2e-dra-gpu-4-21](https://prow.ci.openshift.org/job-history/gs/test-platform-results/logs/periodic-ci-openshift-kueue-operator-main-test-e2e-dra-gpu-4-21) (Problems 1, 2)
  - [periodic-ci-openshift-kueue-operator-main-test-e2e-downstream-4-22](https://prow.ci.openshift.org/job-history/gs/test-platform-results/logs/periodic-ci-openshift-kueue-operator-main-test-e2e-downstream-4-22) (Problem 3)

## Summary by CodeRabbit

* **Tests**
  * Enhanced GPU device type detection validation in Dynamic Resource Allocation Extended Resources scenarios
  * Strengthened test configuration cleanup procedures to prevent interference between test suites
  * Added verification checks for resource reservations and device-class mapping reconciliation
  * Improved ResourceSlice validation before GPU resource counting

### DIFF
--- a/test/e2e/e2e_dra_extended_resources_test.go
+++ b/test/e2e/e2e_dra_extended_resources_test.go
@@ -37,6 +37,7 @@ const (
 	erTestNamespacePrefix = "kueue-dra-er-test-"
 	erLocalQueueName      = "er-queue"
 	extendedResourceName  = "nvidia.com/gpu"
+	gpuDeviceType         = "gpu"
 )
 
 var _ = Describe("DRA Extended Resources", Label("operator", "dra", "dra-extended-resources"), Ordered, func() {
@@ -88,7 +89,7 @@ var _ = Describe("DRA Extended Resources", Label("operator", "dra", "dra-extende
 						// with value "gpu" for full GPU devices. MIG slices have different type values.
 						// See: device.attributes['gpu.nvidia.com'].type in DeviceClass CEL selectors.
 						if typeAttr, ok := d.Attributes["type"]; ok {
-							if typeAttr.StringValue != nil && *typeAttr.StringValue == "gpu" {
+							if typeAttr.StringValue != nil && *typeAttr.StringValue == gpuDeviceType {
 								gpusPerNode[*s.Spec.NodeName]++
 							}
 						}
@@ -116,9 +117,11 @@ var _ = Describe("DRA Extended Resources", Label("operator", "dra", "dra-extende
 		Expect(err).ToNot(HaveOccurred())
 		initialKueueInstance = kueueInstance.DeepCopy()
 
+		// Clear deviceClassMappings if set by a previous test suite
+		kueueInstance.Spec.Config.Resources.DeviceClassMappings = nil
 		applyKueueConfig(ctx, kueueInstance.Spec.Config, kubeClient)
 
-		// Wait for DRA feature gates to be enabled in Kueue config
+		// Wait for DRA feature gates to be enabled and deviceClassMappings to be cleared
 		Eventually(func() error {
 			configMap, err := kubeClient.CoreV1().ConfigMaps(testutils.OperatorNamespace).Get(ctx, "kueue-manager-config", metav1.GetOptions{})
 			if err != nil {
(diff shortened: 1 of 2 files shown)


anahas-redhat:
@sohankunkerkar @kannon92 can you PTAL?
Thank you!

test/e2e/e2e_dra_extended_resources_test.go
```diff
@@ -116,9 +117,11 @@ var _ = Describe("DRA Extended Resources", Label("operator", "dra", "dra-extende
 		Expect(err).ToNot(HaveOccurred())
 		initialKueueInstance = kueueInstance.DeepCopy()
 
+		// Clear deviceClassMappings if set by a previous test suite
```
kannon92, line 120: Rather than patching this test, can we fix the test that pollutes this?

It would be better to have our tests be idempotent.

anahas-redhat: @kannon92 I see your point. So, when DRA suite runs, these two files e2e_dra_extended_resources_test.go and e2e_dra_test.go are executed. 

Technically, a "clear deviceClassMappings" would be needed on e2e_dra_test.go. However, what this file is doing is checking if deviceClassMappings is already configured — if not, it adds the mapping.
The actual deviceClassMappings configuration is actually coming from default Kueue CR [test/e2e/bindata/assets/08_kueue_default.yaml](https://github.com/openshift/kueue-operator/blob/main/test/e2e/bindata/assets/08_kueue_default.yaml).

```
  deviceClassMappings:
  - name: gpu
    deviceClassNames:
    - gpu.example.com
```

  So even if the DRA test's AfterAll perfectly restores the original config, deviceClassMappings is still active (with gpu.example.com), and  Kueue still does resource name remapping. The ER test needs deviceClassMappings to be nil for nvidia.com/gpu to work without translation.

  I see a few options:

  1. Keep the clear in the ER test's BeforeAll (current approach) — each test ensures its own preconditions.
  2. Add an explicit deviceClassMappings = nil in e2e_dra_test.go's AfterAll — so it actively clears the mappings regardless of where they came from.

What do you think?

kannon92: Could we do both 1 and 2?

It sounds like we need to unset the existing setting but also any change should be reset back to normal.

kannon92:
@coderabbit review

kannon92 (commented):
/lgtm
/approve

PR #1888, opened 2026-05-14
add tas downstream tests

kannon92:
Turns out the upstream tests hardcode kind nodes so it would be quite difficult to adapt the upstream tests.

So i went with a downstream.

## Summary by CodeRabbit

* **Tests**
  * Added a comprehensive end-to-end test suite for Topology-Aware Scheduling covering batch jobs, job sets, individual pods, pod groups, and datacenter multi-level scenarios; verifies admission results and topology assignments.
  * Extended test utilities for creating/cleaning topologies and resource flavors, labeling/cleaning nodes, and polling workloads to assert topology details.

* **Chores**
  * Added a CI make target to run TAS-labeled e2e tests and produce JUnit results.

### DIFF
--- a/test/e2e/e2e_tas_test.go
+++ b/test/e2e/e2e_tas_test.go
@@ -1,0 +1,821 @@
+/*
+Copyright 2025.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package e2e
+
+import (
+	"context"
+	"fmt"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	"github.com/openshift/kueue-operator/test/e2e/testutils"
+	appsv1 "k8s.io/api/apps/v1"
+	batchv1 "k8s.io/api/batch/v1"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/utils/ptr"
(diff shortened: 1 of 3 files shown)


test/e2e/e2e_tas_test.go
```diff
@@ -0,0 +1,933 @@
...
+var _ = Describe("TopologyAwareScheduling", Label("tas"), func() {
+
+	When("Creating a Job requesting TAS", func() {
+		var (
+			namespace  *corev1.Namespace
+			topology   *kueuev1beta2.Topology
+			localQueue *kueuev1beta2.LocalQueue
+		)
+
+		BeforeEach(func(ctx context.Context) {
```
MaysaMacedo, line 135: This `BeforeEach` and `JustAfterEach` looks like the same from most of the other ginkgo `When`. Perhaps we could combine those `When` conditions that have same `BeforeEach` into one `When` to avoid code replication. WDYT?

kannon92: Good catch, done — combined the Job/JobSet/Pod/Deployment `When` blocks (which all shared the same hostname-topology `BeforeEach`/`JustAfterEach`) into a single `When("Creating a workload requesting TAS with hostname topology", ...)` block with separate `It`s. Pushed in 2ec0a0f0.

test/e2e/e2e_tas_test.go
```diff
@@ -0,0 +1,823 @@
...
+		g.Expect(blockValues).To(HaveLen(1),
+			fmt.Sprintf("expected all domains to share one block value, got: %v", blockValues))
+	}, testutils.OperatorReadyTime, testutils.OperatorPoll).Should(Succeed())
+}
+
+// verifyWorkloadAdmittedWithTopology checks that a workload for the given job UID
+// is admitted and has TopologyAssignment set on the expected number of PodSetAssignments.
+func verifyWorkloadAdmittedWithTopology(ctx context.Context, namespace, jobUID string, expectedPodSets int, topologyName string) {
+	kueueClient := clients.UpstreamKueueClient
+	_ = topologyName
```
MaysaMacedo: nit: Is the `topologyName` parameter meant to be used anywhere in this function? if not, maybe it could be removed?

kannon92: Good catch — it wasn't used for anything, so I removed the parameter entirely (and updated the call sites) rather than keeping a discarded arg. Pushed in dbe19d61.

MaysaMacedo (approved):
/lgtm

PR #1899, opened 2026-05-18
Remove Default option for AdmissionFairSharing

MaysaMacedo:
When specifying the Default configuration the
field UsageHalfLifeTimeSeconds has the value 0
set as default. Setting this value disables the
feature of Admission Fair Sharing. Following
discussion on kubernetes-sigs/kueue, we're now
making this field required and with value bigger
than 0.

## Summary by CodeRabbit

* **Refactor**
  * AdmissionFairSharing simplified: fair-sharing fields (usageHalfLifeTimeSeconds, usageSamplingIntervalSeconds, resourceWeights) are now top-level under admissionFairSharing instead of nested.
* **Behavior**
  * usageHalfLifeTimeSeconds is now required; its description no longer states omission resets usage to 0.
* **Tests**
  * Validation and config-generation tests updated to use the new top-level fields and expectations.


kannon92 (commented):
/lgtm
/approve

PR #1902, opened 2026-05-19
OCPKUEUE-584: Add e2e test for Admission Fair Sharing with weighted LocalQueues

anahas-redhat:
[OCPKUEUE-584](https://redhat.atlassian.net/browse/OCPKUEUE-584): Add e2e test for Admission Fair Sharing with weighted LocalQueues

Validates that Kueue's Admission Fair Sharing correctly prioritizes
pending workloads based on LocalQueue weight values. The test configures
AFS with usageHalfLifeTimeSeconds=60 and usageSamplingIntervalSeconds=5,
creates two LocalQueues (weight=1 "heavy" and weight=2 "light"), fills
the ClusterQueue quota, then verifies that when quota frees up, the
higher-weight queue's workload is admitted first while the lower-weight
queue's workload remains suspended until additional quota becomes available.

Also adds WithFairSharingWeight and WithAdmissionScope helpers to the
test utilities.

## Summary by CodeRabbit

* **Tests**
  * Added an end-to-end test for Admission Fair Sharing that verifies admission transitions when quota is freed.
  * Added a helper to create long-running suspended jobs and fluent methods to configure admission scope and fair-sharing weight for queues.
  * Extended polling/timeouts for longer "consistently" checks to improve e2e test stability.

PR #1917, opened 2026-05-20
Add test to verify usagehalfLifeTime from admissionFairSharing

MaysaMacedo:
Assisted with: Claude

## Summary by CodeRabbit

* **Tests**
  * Added an ordered end-to-end scenario "Job is created and deleted" that validates fair-sharing admission: enables AdmissionFairSharing with configured half-life and sampling interval, creates required resources and a suspended job, waits for admission, samples CPU consumption until ≥0.9 cores, deletes the job, and verifies consumption decays to 40–60% of pre-delete within one half-life. Updated copyright year.

### DIFF
--- a/test/e2e/e2e_admission_fair_sharing_test.go
+++ b/test/e2e/e2e_admission_fair_sharing_test.go
@@ -1,5 +1,5 @@
 /*
-Copyright 2025.
+Copyright 2026.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


test/e2e/e2e_admission_fair_sharing_test.go
```diff
@@ -0,0 +1,226 @@
...
+				g.Expect(lq.Status.FairSharing).NotTo(BeNil())
+				g.Expect(lq.Status.FairSharing.AdmissionFairSharingStatus).NotTo(BeNil())
+				cpu := lq.Status.FairSharing.AdmissionFairSharingStatus.ConsumedResources[corev1.ResourceCPU]
+				current := cpu.AsApproximateFloat64()
+				klog.Info("consumedResources", "cpu", current)
+				g.Expect(current).Should(BeNumerically(">", lowerBound),
+					"consumedResources decayed too fast (floor %.3f cores)", lowerBound)
+				g.Expect(current).Should(BeNumerically("<", upperBound),
+					"consumedResources did not decay enough (ceiling %.3f cores)", upperBound)
+			}, afsHalfLifeSeconds, time.Duration(afsSamplingSeconds)*time.Second).Should(Succeed(),
```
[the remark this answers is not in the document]
MaysaMacedo: this is now fixed.

test/e2e/e2e_admission_fair_sharing_test.go
```diff
@@ -0,0 +1,221 @@
...
+	"k8s.io/utils/ptr"
+	kueuev1beta2 "sigs.k8s.io/kueue/apis/kueue/v1beta2"
+)
+
+const (
+	afsHalfLifeSeconds = 60
+	afsSamplingSeconds = 5
+)
+
+var _ = Describe("Admission Fair Sharing", Label("operator", "admissionfairsharing"), Ordered, func() {
```
anahas-redhat: Can this label be "admission-fair-sharing"? It would follow patterns like "managed-jobs-ns-selector", "local-queue-default", etc.

test/e2e/e2e_admission_fair_sharing_test.go
```diff
@@ -0,0 +1,221 @@
...
+				ObjectMeta: metav1.ObjectMeta{
+					GenerateName: "afs-test-",
+					Labels:       map[string]string{testutils.OpenShiftManagedLabel: "true"},
+				},
+			}, metav1.CreateOptions{})
+			Expect(err).NotTo(HaveOccurred())
+			ns = createdNs
+
+			By("Creating ResourceFlavor")
+			var err2 error
```
anahas-redhat: can't we reuse err?

test/e2e/e2e_admission_fair_sharing_test.go
```diff
@@ -0,0 +1,221 @@
...
+			Expect(err2).NotTo(HaveOccurred())
+
+			By("Creating LocalQueue lq-a bound to the ClusterQueue")
+			lqA, _, err2 = testutils.NewLocalQueue(ns.Name, "lq-a").
+				WithClusterQueue(cq.Name).
+				CreateWithObject(ctx, clients.UpstreamKueueClient)
+			Expect(err2).NotTo(HaveOccurred())
+		})
+
+		AfterEach(func(ctx context.Context) {
```
anahas-redhat, line 617: I know that when the namespace is deleted, the localqueue is deleted as well. However, it seems to be a pattern to delete the localqueue as well, like: [e2e_operator_test.go](https://github.com/openshift/kueue-operator/blob/main/test/e2e/e2e_operator_test.go#L492) and [e2e_visibility_on_demand_test.go](https://github.com/openshift/kueue-operator/blob/main/test/e2e/e2e_visibility_on_demand_test.go#L93), for example. WDYT?

test/e2e/e2e_admission_fair_sharing_test.go
```diff
@@ -0,0 +1,221 @@
...
+			deleteNamespace(ctx, ns)
```
anahas-redhat, line 621: I think what coderabbit is saying here make sense. The AfterEach calls cleanupCQ() and cleanupRF() directly, which will panic if BeforeEach fails. We can use the same pattern of [local_queue_defaulting_test.go](https://github.com/openshift/kueue-operator/blob/main/test/e2e/e2e_local_queue_defaulting_test.go#L312), for example.

test/e2e/e2e_admission_fair_sharing_test.go
```diff
@@ -0,0 +1,221 @@
...
+				g.Expect(current).Should(BeNumerically("<", upperBound),
+					"consumedResources did not decay enough (ceiling %.3f cores)", upperBound)
+			}, afsHalfLifeSeconds*time.Second, afsSamplingSeconds*time.Second).Should(Succeed(),
+				"consumedResources did not decay to ~50%% of pre-delete value after one usageHalfLifeTimeSeconds")
+			GinkgoWriter.Printf("Result: consumedResources.cpu = %.3f cores\n", current)
+		})
+	})
+})
+
+func afsJob(name, namespace, queueName string) *batchv1.Job {
```
anahas-redhat: Can we leverage this function  [newLongRunningJob(name, namespace, queueName, cpu, memory)](https://github.com/openshift/kueue-operator/pull/1902/changes#diff-09d00e74686d9dc8e40c4c874417afe1a5ef701618682b33fe699d36fc075cb5R207) on PR #1902 ?
They are pretty similar but afsJob seems to have a hardcoded CPU. Reusing newLongRunningJob avoids two duplicate job builders in the same file (will also help on the deletion, below).

test/e2e/e2e_admission_fair_sharing_test.go
```diff
@@ -0,0 +1,221 @@
...
+							},
+						},
+					},
+				},
+			},
+		},
+	}
+}
+
+func afsDeleteJob(ctx context.Context, namespace, name string) {
```
anahas-redhat: If we use the function above (newLongRunningJob) there's no need for a custom deletion helper. The same inline delete pattern from PR #1902 works (a Delete with DeletePropagationBackground)

anahas-redhat (commented):
Hi @MaysaMacedo, I'll add below my comments for this PR review. Let me know your thoughts.

MaysaMacedo:
/hold 
for the API change PR to get in first.

MaysaMacedo:
/hold cancel

I had to rebase and adapted the BeforeAll to use the structure that was already present and could be re-used.

PR #1939, opened 2026-05-27
OCPKUEUE-585 Add e2e test for Admission Fair Sharing with resourceWeights

anahas-redhat:
**Testing scenario:**
Validates that configuring resourceWeights (cpu: 10, memory: 0) in the Kueue operator correctly influences fair sharing scoring. 
The test verifies that when two LocalQueues have equal weights but different CPU usage, the queue with lower weighted CPU usage wins admission when quota frees up. 
Memory weight is set to 0 to work around upstream bug kubernetes-sigs/kueue#10434 (OCPBUGS-85113). 

**What is the upstream issue:**
The issue is an upstream Kueue bug ([kubernetes-sigs/kueue#10434](https://github.com/kubernetes-sigs/kueue/issues/10434)) in how fair sharing scores are calculated. When Kueue computes a fair sharing score for a ClusterQueue, it multiplies each resource's usage by its weight. The problem is that memory usage is counted in raw bytes rather than being normalized to a human-meaningful unit like GiB. So, it would work like this:
- A workload requesting 1 CPU contributes 1 × weight to the score.
- A workload requesting 1Gi memory contributes 1,073,741,824 × weight to the score (because 1Gi = 1,073,741,824 bytes).
So even with equal weights (cpu: 1, memory: 1), the memory term utterly dominates the score — it's over a billion times larger than the CPU term. The CPU weight becomes meaningless.

Given this upstream limitation, isolating CPU as the sole scoring factor by setting memory weight to 0 seems to be a reliable approach for deterministic fair sharing behavior for now.

## Summary by CodeRabbit

* **Tests**
  * Added an end-to-end test for Admission Fair Sharing that verifies resource-weight-driven fair-sharing scoring. The scenario exercises two queues with differing CPU usage profiles, saturates capacity with long-running jobs, introduces pending CPU-consuming jobs, frees capacity, and asserts that admission favors the queue with a lower CPU usage score while the other remains suspended.

### DIFF
--- a/test/e2e/e2e_admission_fair_sharing_test.go
+++ b/test/e2e/e2e_admission_fair_sharing_test.go
@@ -420,9 +420,153 @@ var _ = Describe("Admission Fair Sharing", Label("admission-fair-sharing"), Orde
 		})
 	})
 
+	// Upstream bug: kubernetes-sigs/kueue#10434 (OCPBUGS-85113)
+	// Memory resourceWeights are counted in raw bytes instead of GiB, so non-zero
+	// memory weights produce unintuitive scores. Setting memory weight to "0"
+	// sidesteps this and isolates CPU as the sole scoring factor.
+	When("resourceWeights customizes which resources drive fair sharing scoring", func() {
+
+		It("should deprioritize the queue with higher weighted CPU usage when cpu weight is 10 and memory weight is 0", func(ctx context.Context) {
+			By("Overriding Kueue config with resourceWeights (cpu: 10, memory: 0) and fast decay")
+			enableAdmissionFairSharing(ctx, 10, 1,
+				ssv1.ResourceWeight{Name: "cpu", Weight: "10"},
+				ssv1.ResourceWeight{Name: "memory", Weight: "0"},
+			)
+
+			By("Creating Resource Flavor")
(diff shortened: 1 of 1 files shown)


test/e2e/e2e_admission_fair_sharing_test.go
```diff
@@ -420,9 +420,153 @@ var _ = Describe("Admission Fair Sharing", Label("admission-fair-sharing"), Orde
 		})
 	})
 
+	// Upstream bug: kubernetes-sigs/kueue#10434 (OCPBUGS-85113)
```
kannon92, line 423: Any reason why we can't add this e2e test upstream? Or add it as an integration test?

anahas-redhat: I guess it would be the same reason we did not create the other tests upstream. I think the goal is to  validate the operator's translation of ResourceWeights from the Kueue CR into the kueue-manager-config ConfigMap, and that Kueue picks it up correctly on a real cluster (operator CR -> ConfigMap -> behavior end-to-end).

Once the upstream bug (kubernetes-sigs/kueue#10434) is fixed, we could contribute integration tests upstream that cover resourceWeights with both CPU and memory weights without the workaround. 

WDYT?

kannon92 (commented):
/lgtm

PR #1978, opened 2026-06-02
Add API fields for DRA integration and partitionable devices support

PannagaRao:

## Summary by CodeRabbit

* **New Features**
  * Device-class mappings gain an optional single-entry "sources" configuration for resource accounting, supporting counter-based sources with driver, counter name, and CEL-based device selection.
  * Operator and generated controller configuration now include a partitionable-devices feature gate when enabled.

* **Tests**
  * ConfigMap unit tests updated to include the new partitionable-devices flag.


pkg/apis/kueueoperator/v1/types.go
```diff
@@ -2,6 +2,7 @@ package v1
 
 import (
 	operatorv1 "github.com/openshift/api/operator/v1"
+	resourcev1 "k8s.io/api/resource/v1"
```
sohankunkerkar: The operator API types should stay decoupled from upstream k8s DRA types.

pkg/apis/kueueoperator/v1/types.go
```diff
@@ -443,6 +444,58 @@ type DeviceClassMapping struct {
...
+
+	// deviceSelector scopes which devices are eligible for counter-based
+	// quota accounting. Typically matches a GPU model (e.g., productName)
+	// so all partition profiles on that model share one quota pool.
+	// Per-workload charging is determined by the workload's own
+	// ResourceClaimTemplate selector, which narrows to the requested profile.
+	// The selector is compiled at config load time using the upstream dracel
+	// compiler.
+	// +required
+	DeviceSelector resourcev1.DeviceSelector `json:"deviceSelector"`
```
sohankunkerkar: You can do something like:
```go
  // DeviceSelector selects devices by attributes.                                                                                                                                                                 
  type DeviceSelector struct {                                                                                                                                                                                     
      // cel contains a CEL expression for filtering devices.                                                                                                                                                      
      // +optional                                                                                                                                                                                                 
      CEL *CELDeviceSelector `json:"cel,omitempty"`                                                                                                                                                                
  }                                                                                                       

  // CELDeviceSelector filters devices using a CEL expression.                                            
  type CELDeviceSelector struct {                    
      // expression is a CEL expression that evaluates against device attributes.
      // +required
      Expression string `json:"expression"`                                                               
  }
```

and then convert it to the upstream DeviceSelector struct in configmap.go where the submodule types are already available.

PannagaRao: Fixed.

sohankunkerkar:
@JoelSpeed Is there a timeline for bumping openshift/api to k8s.io/api v0.36? We need sigs.k8s.io/kueue@v0.18.0 in the kueue-operator which requires k8s 0.36, but it conflicts with openshift/api still on 0.35. Want to know if we should wait for the rebase or define the new types locally as a workaround.

kannon92:
> @JoelSpeed Is there a timeline for bumping openshift/api to k8s.io/api v0.36? We need sigs.k8s.io/kueue@v0.18.0 in the kueue-operator which requires k8s 0.36, but it conflicts with openshift/api still on 0.35. Want to know if we should wait for the rebase or define the new types locally as a workaround.

Not sure why Joel is pinged. You can follow tmp-k8s-1-36 in internal slack to find status of the rebase.

sohankunkerkar:
> > @JoelSpeed Is there a timeline for bumping openshift/api to k8s.io/api v0.36? We need sigs.k8s.io/kueue@v0.18.0 in the kueue-operator which requires k8s 0.36, but it conflicts with openshift/api still on 0.35. Want to know if we should wait for the rebase or define the new types locally as a workaround.
> 
> Not sure why Joel is pinged. You can follow tmp-k8s-1-36 in internal slack to find status of the rebase.

Thanks for the pointer to #tmp-k8s-1-36. I pinged Joel so he's aware of this dependency and we can get the API reviews unblocked on this PR irrespective of the rebase timeline.

kannon92:
I followed off ofline with @PannagaRao  and the library-go update to 1.36 is the main thing actually.

pkg/apis/kueueoperator/v1/types.go
```diff
@@ -443,6 +443,74 @@ type DeviceClassMapping struct {
 	// +kubebuilder:validation:MinItems=1
 	// +required
 	DeviceClassNames []DeviceClassName `json:"deviceClassNames,omitempty"`
+
+	// sources configures resource accounting sources for this mapping.
+	// Each source defines how quota is tracked for this DeviceClass.
+	// Currently only counter sources are supported (for partitionable devices).
+	// Extended resource requests that resolve to a DeviceClass with sources
+	// configured are marked inadmissible.
+	// Requires the KueueDRAIntegrationPartitionableDevices feature gate.
```
kannon92: Im not sure this statement is true. 

Users have no way to change this feature gate so is there any point in putting this here? I'd expect us to set the feature gate for users if they specify this.

PannagaRao: We could enable the feature gate is Sources are configured. But without k8s gate being enabled workloads will not be admitted. Do you think we should check for gate and Sources in config before enabling it or just document that it needs k8s gate?

kannon92: Hmm. But I guess that also depends. What should we do if someone sets these APIs without PD enabled?

PannagaRao: Maybe show a degraded state?

everettraven: Presumably this is a feature-gate that is used by the kueue-operator itself to gate logic associated with this functionality?

How would a feature-gate for this operator get enabled by an end-user of OpenShift today?

everettraven: Ah I see. I had forgotten that this operator is a meta-operator for the actual kueue-operator.

If you require specific gates on the kube-apiserver to be enabled for this functionality to work, I'd make sure that the steps to enable them are explicitly documented.

In this case, it looks like the `DRAPartitionableDevices` gate is going to be beta in 1.36 and enabled by default: https://github.com/kubernetes/kubernetes/blob/a94f8bfc3eb5e8cdaf1281ab730dcd56c365323c/pkg/features/kube_features.go#L1322

That means it _should_ be enabled by default in 4.23/5.0 if we don't have any gate overrides in openshift/api.

Unfortunately, OLM-deployed operators have no way of doing the same kind of TPNU gating of CRDs as we do for payload-shipped CRDs.

I can't think of any other way for you to reliably be able to check if the dependent feature is enabled or not without an explicit gate for OpenShift, but it would end up overriding the gate state of the upstream gate for OpenShift clusters.

The only way I can think of is to make sure that any version of the operator shipped with this functionality is only installable on OpenShift clusters that are running K8s 1.36+ (which would be 4.23/5.0+). I _think_ there is a way for OLM-installed operators to specify that - or at least ensure that version is only present in the catalogs meant for that version of OpenShift, but that would be something to confirm with the OLM folks.

pkg/apis/kueueoperator/v1/types.go
```diff
@@ -443,6 +443,74 @@ type DeviceClassMapping struct {
...
+	// +kubebuilder:validation:MaxItems=1
+	// +optional
+	Sources []DeviceClassSourceConfig `json:"sources,omitempty"`
```
everettraven, line 460: It looks like the upstream type doesn't have a documented limitation to the maximum size, where is our max items of 1 coming from here?

Additionally, if the empty value is not valid here we should add a minimum items of 1 to ensure that at least one entry must be provided when this field is set.

pkg/apis/kueueoperator/v1/types.go
```diff
@@ -443,6 +443,74 @@ type DeviceClassMapping struct {
...
+
+// DeviceClassSourceConfig defines a resource accounting source for a DeviceClassMapping.
+// Exactly one of the source types must be set.
+// +kubebuilder:validation:MinProperties=1
+type DeviceClassSourceConfig struct {
+	// counter configures counter-based quota for partitionable devices.
+	// Maps a DRA driver counter to the parent DeviceClassMapping's Kueue quota resource.
+	// +optional
+	Counter *DeviceClassCounterSource `json:"counter,omitempty"`
+}
```
everettraven, line 479: This looks like it would be a good candidate for a discriminated union: https://github.com/openshift/enhancements/blob/master/dev-guide/api-conventions.md#discriminated-unions

pkg/apis/kueueoperator/v1/types.go
```diff
@@ -443,6 +443,74 @@ type DeviceClassMapping struct {
...
+	// entries to track for quota. Must match a counter name published by
+	// the driver in ResourceSlice devices' consumesCounters field.
+	// Counter set names are per-device identifiers (e.g., gpu-0-counter-set,
+	// gpu-1-counter-set), so name matches across all counter sets
+	// for a given driver without requiring one mapping per device.
+	// The total length must not exceed 63 characters.
+	// +kubebuilder:validation:MaxLength=63
+	// +kubebuilder:validation:MinLength=1
+	// +required
+	Name string `json:"name"`
```
everettraven:
```suggestion
	Name string `json:"name,omitempty"`
```

pkg/apis/kueueoperator/v1/types.go
```diff
@@ -443,6 +443,74 @@ type DeviceClassMapping struct {
...
+
+	// driver is the DRA driver name used to filter relevant ResourceSlices.
+	// Must match the spec.driver field on ResourceSlice objects.
+	// The total length must not exceed 253 characters.
+	// +kubebuilder:validation:MaxLength=253
+	// +kubebuilder:validation:MinLength=1
+	// +required
+	Driver string `json:"driver"`
```
everettraven:
```suggestion
	Driver string `json:"driver,omitempty"`
```

pkg/apis/kueueoperator/v1/types.go
```diff
@@ -443,6 +443,74 @@ type DeviceClassMapping struct {
...
+
+	// deviceSelector scopes which devices are eligible for counter-based
+	// quota accounting. Typically matches a GPU model (e.g., productName)
+	// so all partition profiles on that model share one quota pool.
+	// Per-workload charging is determined by the workload's own
+	// ResourceClaimTemplate selector, which narrows to the requested profile.
+	// The selector is compiled at config load time using the upstream dracel
+	// compiler.
+	// +required
+	DeviceSelector DeviceSelector `json:"deviceSelector"`
```
everettraven:
```suggestion
	DeviceSelector DeviceSelector `json:"deviceSelector,omitzero"`
```

pkg/apis/kueueoperator/v1/types.go
```diff
@@ -443,6 +443,74 @@ type DeviceClassMapping struct {
...
+}
+
+// DeviceSelector selects devices by attributes.
+type DeviceSelector struct {
+	// cel contains a CEL expression for filtering devices.
+	// +required
+	CEL *CELDeviceSelector `json:"cel"`
```
everettraven:
```suggestion
	CEL CELDeviceSelector `json:"cel,omitzero"`
```

pkg/apis/kueueoperator/v1/types.go
```diff
@@ -443,6 +443,74 @@ type DeviceClassMapping struct {
...
+}
+
+// CELDeviceSelector filters devices using a CEL expression.
+type CELDeviceSelector struct {
+	// expression is a CEL expression that evaluates against device attributes.
+	// +kubebuilder:validation:MinLength=1
+	// +required
+	Expression string `json:"expression"`
```
everettraven:
```suggestion
	Expression string `json:"expression,omitempty"`
```

pkg/apis/kueueoperator/v1/types.go
```diff
@@ -443,6 +443,74 @@ type DeviceClassMapping struct {
...
+type DeviceSelector struct {
+	// cel contains a CEL expression for filtering devices.
+	// +required
+	CEL *CELDeviceSelector `json:"cel"`
+}
+
+// CELDeviceSelector filters devices using a CEL expression.
+type CELDeviceSelector struct {
+	// expression is a CEL expression that evaluates against device attributes.
+	// +kubebuilder:validation:MinLength=1
```
everettraven, line 553: Please also add a reasonable maximum length. CEL expressions is arbitrary code that is compiled and as such, keeping this unbounded means that end-users can end up writing extremely complex expressions that take a while to compile.

I've implemented fields like this fairly recently and we set the maximum size to 1024 as even deeply nested expensive calls compiled in a reasonably short period of time.

pkg/apis/kueueoperator/v1/types.go
```diff
@@ -443,6 +443,74 @@ type DeviceClassMapping struct {
...
+	// Per-workload charging is determined by the workload's own
+	// ResourceClaimTemplate selector, which narrows to the requested profile.
+	// The selector is compiled at config load time using the upstream dracel
+	// compiler.
+	// +required
+	DeviceSelector DeviceSelector `json:"deviceSelector"`
+}
+
+// DeviceSelector selects devices by attributes.
+type DeviceSelector struct {
```
everettraven, line 530: Do you anticipate this selector ever having additional selector fields beyond `cel`?

pkg/apis/kueueoperator/v1/types.go
kannon92: > I can't think of any other way for you to reliably be able to check if the dependent feature is enabled or not without an explicit gate for OpenShift, but it would end up overriding the gate state of the upstream gate for OpenShift clusters.

We could check OCP version.

> The only way I can think of is to make sure that any version of the operator shipped with this functionality is only installable on OpenShift clusters that are running K8s 1.36+ (which would be 4.23/5.0+). I think there is a way for OLM-installed operators to specify that - or at least ensure that version is only present in the catalogs meant for that version of OpenShift, but that would be something to confirm with the OLM folks.

This is a non-starter for us. Dynamic MIG is one feature of Kueue and we are aiming to support 4.18 and above for as long as EUS is.

So our goal is to avoid detect invalid configurations and move operator to degraded state. This is a pattern we did for DRA features that rely on 4.21. I think we can do something similar here.

The reality is that if Kubernetes itself is not configured for Dynamic MIG, I think whatever Kueue tries to do won't work either so I wouldn't be that worried about it.

pkg/apis/kueueoperator/v1/types.go
```diff
@@ -443,6 +443,74 @@ type DeviceClassMapping struct {
...
+	// Maps a DRA driver counter to the parent DeviceClassMapping's Kueue quota resource.
+	// +optional
+	Counter *DeviceClassCounterSource `json:"counter,omitempty"`
+}
+
+// DeviceClassCounterSource identifies where to read counter data from and which counter to track.
+type DeviceClassCounterSource struct {
+	// name is the counter name within the device's consumesCounters
+	// entries to track for quota. Must match a counter name published by
+	// the driver in ResourceSlice devices' consumesCounters field.
```
everettraven, line 492: Is there a particular format to counter names that would exist in this field that we can enforce here?

pkg/apis/kueueoperator/v1/types.go
```diff
@@ -443,6 +443,74 @@ type DeviceClassMapping struct {
...
+	// gpu-1-counter-set), so name matches across all counter sets
+	// for a given driver without requiring one mapping per device.
+	// The total length must not exceed 63 characters.
+	// +kubebuilder:validation:MaxLength=63
+	// +kubebuilder:validation:MinLength=1
+	// +required
+	Name string `json:"name"`
+
+	// driver is the DRA driver name used to filter relevant ResourceSlices.
+	// Must match the spec.driver field on ResourceSlice objects.
```
everettraven, line 506: IIRC driver names have a particular naming format (drawing a blank on what it is). We should ensure we include any formatting constraints as explicit validations here.

pkg/apis/kueueoperator/v1/types.go
PannagaRao: @kannon92 So to confirm — we enable the kueue gate when sources are configured AND K8s gate is detected, and degrade when sources are set without the k8s gate.Does that sound good?

kannon92: I think so.

everettraven: How are you checking for the gate enablement? For k8s-only gates, there isn't really a way for you to identify if they are enabled our not unless you know the default enablements states for the current k8s version.

kannon92: Hmm good point.

https://github.com/openshift/kueue-operator/blob/main/pkg/operator/target_config_reconciler.go#L359

This is what we have been doing.

DRA we check if the API exists.

ExtendedResources we check for CustomNoUpgrade.

We could do the same for PartitionableDevices. This logic may need to be rethough for 4.23/5.0 (as these features will be enabled by default so CustomNoUpgrade will not be spcified).

pkg/apis/kueueoperator/v1/types.go
PannagaRao: Our DeviceSelector mirrors upstream k8s which has only `cel` today. I made it required as it is currently the only type. Should I use discriminated union here as well ?

pkg/apis/kueueoperator/v1/types.go
PannagaRao: Upstream Kueue doesn't set its own limit. @sohankunkerkar can we also use 1024 as maximum length?

pkg/apis/kueueoperator/v1/types.go
PannagaRao: MaxItems=1 comes from upstream's runtime validation in pkg/config/validation.go. We enforce the same at CRD level. Will add MinItems=1 as well.

pkg/apis/kueueoperator/v1/types.go
sohankunkerkar: +1 to this

pkg/apis/kueueoperator/v1/types.go
sohankunkerkar: Actually, upstream K8s [enforces](https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/api/resource/v1beta2/types.go#L1553) CELSelectorExpressionMaxLength = 10 * 1024 (10 KiB) in resource/v1beta2/types.go, validated at API admission time. I think we should match that if possible. Practically, real device selector expressions are well under 1 KiB, but matching upstream avoids rejecting expressions that K8s itself would accept.

pkg/apis/kueueoperator/v1/types.go
PannagaRao: Will add a DNS label validation for counter names.

pkg/apis/kueueoperator/v1/types.go
PannagaRao: We could add DNS subdomain validation. Does that sound good?

pkg/apis/kueueoperator/v1/types.go
everettraven: I would recommend making this a DU as well in the case that upstream adds support for more than CEL here. DU is a better pattern for supporting different "types" of configuration than co-located fields, especially for readers of an API and is a pattern that most OpenShift users will be familiar with.

pkg/apis/kueueoperator/v1/types.go
```diff
@@ -443,6 +443,114 @@ type DeviceClassMapping struct {
...
+	// Currently only counter sources are supported (for partitionable devices).
+	// Extended resource requests that resolve to a DeviceClass with sources
+	// configured are marked inadmissible.
+	// The operator automatically enables the required kueue feature gate when
+	// sources are configured and the Kubernetes DRAPartitionableDevices
+	// feature gate is enabled on the cluster.
+	// +kubebuilder:validation:MinItems=1
+	// +kubebuilder:validation:MaxItems=1
+	// +optional
+	Sources []DeviceClassSourceConfig `json:"sources,omitempty"`
```
everettraven, line 460: Even though you only allow 1 entry in this list, you should also add the appropriate SSA tags like `+listType` here so you don't get locked into the default one (`atomic`) unintentionally.

pkg/apis/kueueoperator/v1/types.go
```diff
@@ -443,6 +443,114 @@ type DeviceClassMapping struct {
...
+	// Counter uses DRA ConsumesCounters data from ResourceSlices to compute quota charges.
+	// +unionDiscriminator
+	// +required
+	Type DeviceClassSourceType `json:"type"`
+
+	// counter configures counter-based quota for partitionable devices.
+	// Maps a DRA driver counter to the parent DeviceClassMapping's Kueue quota resource.
+	// counter is required when type is Counter, and forbidden otherwise.
+	// +optional
+	Counter *DeviceClassCounterSource `json:"counter,omitempty"`
```
everettraven: Can be:

```suggestion
	Counter DeviceClassCounterSource `json:"counter,omitzero"`
```

pkg/apis/kueueoperator/v1/types.go
```diff
@@ -443,6 +443,114 @@ type DeviceClassMapping struct {
...
+	// the driver in ResourceSlice devices' consumesCounters field.
+	// Counter set names are per-device identifiers (e.g., gpu-0-counter-set,
+	// gpu-1-counter-set), so name matches across all counter sets
+	// for a given driver without requiring one mapping per device.
+	// Must be a valid DNS label consisting of lowercase alphanumeric characters
+	// and hyphens, of at most 63 characters in length.
+	// Must start and end with an alphanumeric character.
+	// +kubebuilder:validation:MaxLength=63
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:XValidation:rule="self.matches(r'^[a-z0-9]([-a-z0-9]*[a-z0-9])?$')",message="must be a valid DNS label (e.g., 'memory')"
```
everettraven: For DNS labels, I think you can actually use the built-in `format` library: https://kubernetes.io/docs/reference/using-api/cel/#kubernetes-format-library

pkg/apis/kueueoperator/v1/types.go
```diff
@@ -443,6 +443,114 @@ type DeviceClassMapping struct {
...
+	Name string `json:"name,omitempty"`
+
+	// driver is the DRA driver name used to filter relevant ResourceSlices.
+	// Must match the spec.driver field on ResourceSlice objects.
+	// Must be a valid DNS subdomain consisting of lowercase alphanumeric characters,
+	// hyphens and periods, of at most 253 characters in length.
+	// Each period separated segment must start and end with an alphanumeric character.
+	// +kubebuilder:validation:MaxLength=253
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:XValidation:rule="self.matches(r'^[a-z0-9]([-a-z0-9]*[a-z0-9])?(\\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$')",message="must be a valid DNS subdomain (e.g., 'gpu.nvidia.com')"
```
everettraven: Similarly, there should be a DNS subdomain format utility you can use here.

pkg/apis/kueueoperator/v1/types.go
everettraven: That is a pretty large expression. At just 4KiB we found that it took on the scale of minutes to compile a deeply nested expression at that size.

It is up to you what limit you want to enforce here, but I'd recommend taking worst-case compilation time into account. Especially if changes in the configuration here trigger rollouts of kueue itself (i.e a significantly complex expression could likely make kueue unavailable for some amount of time).

everettraven (commented):
Reviewed this again on Friday but looks like I forgot to post it 🤦

pkg/apis/kueueoperator/v1/types.go
PannagaRao: Thanks for pointing that out. Fixed it

pkg/apis/kueueoperator/v1/types.go
PannagaRao: Done

pkg/apis/kueueoperator/v1/types.go
PannagaRao: I've added `+listType=map` with `+listMapKey=type` on Sources, since type uniquely identifies each entry. Please let me know if that sounds good.

pkg/apis/kueueoperator/v1/types.go
sohankunkerkar: @everettraven Thanks for flagging this. I ran a quick benchmark against the dracel compile path used here with deeply nested expressions up to 10 KiB and saw about 95ms in the worst case, so we weren’t able to reproduce the “minutes at 4 KiB” behavior on this code path. Config changes here do trigger a Kueue rollout, but the CEL compilation happens during pod startup and this cost appears negligible relative to normal startup time. Any thoughts here?

PannagaRao: Added a validation of MaxLength=1024. @everettraven could you please re-review. Thanks!

pkg/apis/kueueoperator/v1/types.go
```diff
@@ -443,6 +443,117 @@ type DeviceClassMapping struct {
...
+	// the driver in ResourceSlice devices' consumesCounters field.
+	// Counter set names are per-device identifiers (e.g., gpu-0-counter-set,
+	// gpu-1-counter-set), so name matches across all counter sets
+	// for a given driver without requiring one mapping per device.
+	// Must be a valid DNS label consisting of lowercase alphanumeric characters
+	// and hyphens, of at most 63 characters in length.
+	// Must start and end with an alphanumeric character.
+	// +kubebuilder:validation:MaxLength=63
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:XValidation:rule="!format.dns1123Label().validate(self).hasValue()",message="must be a valid DNS label (e.g., 'memory')"
```
everettraven: Nit: Have the error message on a failed validation match the GoDoc sentence that explains what makes up a valid DNS label

pkg/apis/kueueoperator/v1/types.go
```diff
@@ -443,6 +443,117 @@ type DeviceClassMapping struct {
...
+	// type selects the device selector type.
+	// CEL uses a CEL expression to filter devices by attributes.
+	// +unionDiscriminator
+	// +required
+	Type DeviceSelectorType `json:"type"`
+
+	// cel contains a CEL expression for filtering devices.
+	// cel is required when type is CEL, and forbidden otherwise.
+	// +optional
+	CEL *CELDeviceSelector `json:"cel,omitempty"`
```
everettraven: This doesn't look like it needs to be a pointer because it has a required child field:

```suggestion
	CEL CELDeviceSelector `json:"cel,omitzero"`
```

PannagaRao: Done

pkg/apis/kueueoperator/v1/types.go
PannagaRao: Done. I have changed it to match go docs. PTAL!

pkg/apis/kueueoperator/v1/types.go
```diff
@@ -443,6 +443,117 @@ type DeviceClassMapping struct {
...
+	Name string `json:"name,omitempty"`
+
+	// driver is the DRA driver name used to filter relevant ResourceSlices.
+	// Must match the spec.driver field on ResourceSlice objects.
+	// Must be a valid DNS subdomain consisting of lowercase alphanumeric characters,
+	// hyphens and periods, of at most 253 characters in length.
+	// Each period separated segment must start and end with an alphanumeric character.
+	// +kubebuilder:validation:MaxLength=253
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:XValidation:rule="!format.dns1123Subdomain().validate(self).hasValue()",message="must be a valid DNS subdomain (e.g., 'gpu.nvidia.com')"
```
everettraven: Same message nit here.

everettraven (approved):
Small nit that could be addressed later as well.

API structure and validations look good to me.

sohankunkerkar (approved):
/retest
/lgtm

PannagaRao:
E2Es for this feature are added in https://github.com/openshift/kueue-operator/pull/1990. Manually verified that they run on a cluster with A100 GPU.

```
[ReportAfterSuite] Autogenerated ReportAfterSuite for --junit-report
autogenerated by Ginkgo
[ReportAfterSuite] PASSED [0.013 seconds]
------------------------------

Ran 14 of 108 Specs in 623.486 seconds
SUCCESS! -- 14 Passed | 0 Failed | 0 Pending | 94 Skipped
PASS
```
cc: @sohankunkerkar @kannon92

PR #1990, opened 2026-06-09
Implement E2Es for Partitionable Devices

PannagaRao:
This PR has a dependency on https://github.com/openshift/kueue-operator/pull/2031

## Release Notes

* **Tests**
  * Added comprehensive e2e test suite for DRA Partitionable Devices, covering device-count charging, multi-device claims, and preemption scenarios.
  * Added test utilities for DRA workload creation and validation.

### DIFF
--- a/test/e2e/e2e_dra_partitionable_devices_test.go
+++ b/test/e2e/e2e_dra_partitionable_devices_test.go
@@ -1,0 +1,727 @@
+/*
+Copyright 2026.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package e2e
+
+import (
+	"context"
+	"fmt"
+	"strings"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	configv1 "github.com/openshift/api/config/v1"
+	configclientv1 "github.com/openshift/client-go/config/clientset/versioned/typed/config/v1"
+	ssv1 "github.com/openshift/kueue-operator/pkg/apis/kueueoperator/v1"
+	"github.com/openshift/kueue-operator/test/e2e/testutils"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/utils/ptr"
+	kueuev1beta2 "sigs.k8s.io/kueue/apis/kueue/v1beta2"
+)
+
+const (
+	pdDriverName          = "gpu.nvidia.com"
+	pdMigDeviceClass      = "mig.nvidia.com"
+	pdGpuDeviceClass      = "gpu.nvidia.com"
+	pdLogicalResource     = "gpu.memory"
+	pdTestNamespacePrefix = "kueue-pd-test-"
+	pdLocalQueueName      = "pd-test-queue"
+)
+
+func verifyWorkloadGPUMemory(ctx context.Context, namespace, jobUID string, expectedBytes int64) {
+	kueueClient := clients.UpstreamKueueClient
+	Eventually(func(g Gomega) {
+		wlList, err := kueueClient.KueueV1beta2().Workloads(namespace).List(ctx, metav1.ListOptions{
+			LabelSelector: fmt.Sprintf("kueue.x-k8s.io/job-uid=%s", jobUID),
+		})
+		g.Expect(err).NotTo(HaveOccurred())
+		g.Expect(wlList.Items).NotTo(BeEmpty())
+
+		wl := wlList.Items[0]
+		g.Expect(wl.Status.Admission).NotTo(BeNil())
+		g.Expect(wl.Status.Admission.PodSetAssignments).To(HaveLen(1))
+
+		assignment := wl.Status.Admission.PodSetAssignments[0]
+		g.Expect(assignment.ResourceUsage).To(HaveKey(corev1.ResourceName(pdLogicalResource)))
+		memUsage := assignment.ResourceUsage[corev1.ResourceName(pdLogicalResource)]
+		g.Expect(memUsage.Value()).To(Equal(expectedBytes), "expected gpu.memory=%d, got=%d", expectedBytes, memUsage.Value())
+	}, testutils.OperatorReadyTime, testutils.OperatorPoll).Should(Succeed())
+}
+
+func verifyWorkloadInadmissible(ctx context.Context, namespace, jobUID string) {
+	kueueClient := clients.UpstreamKueueClient
+	Eventually(func(g Gomega) {
+		wlList, err := kueueClient.KueueV1beta2().Workloads(namespace).List(ctx, metav1.ListOptions{
+			LabelSelector: fmt.Sprintf("kueue.x-k8s.io/job-uid=%s", jobUID),
+		})
+		g.Expect(err).NotTo(HaveOccurred())
+		g.Expect(wlList.Items).NotTo(BeEmpty())
+
+		wl := wlList.Items[0]
+		for _, c := range wl.Status.Conditions {
+			if c.Type == string(kueuev1beta2.WorkloadQuotaReserved) {
+				g.Expect(c.Status).To(Equal(metav1.ConditionFalse))
+				g.Expect(c.Reason).To(Equal("Inadmissible"))
+				return
+			}
+		}
+		g.Expect(false).To(BeTrue(), "QuotaReserved condition not found")
+	}, testutils.OperatorReadyTime, testutils.OperatorPoll).Should(Succeed())
+}
+
+func verifyCQGPUMemory(ctx context.Context, cqName string, expectedBytes int64) {
+	kueueClient := clients.UpstreamKueueClient
+	Eventually(func(g Gomega) {
+		cq, err := kueueClient.KueueV1beta2().ClusterQueues().Get(ctx, cqName, metav1.GetOptions{})
+		g.Expect(err).NotTo(HaveOccurred())
+		for _, flavorUsage := range cq.Status.FlavorsReservation {
+			for _, res := range flavorUsage.Resources {
+				if res.Name == corev1.ResourceName(pdLogicalResource) {
+					g.Expect(res.Total.Value()).To(Equal(expectedBytes))
+					return
+				}
+			}
+		}
+		g.Expect(false).To(BeTrue(), "gpu.memory not found in CQ status")
+	}, testutils.OperatorReadyTime, testutils.OperatorPoll).Should(Succeed())
+}
+
+var _ = Describe("DRA Partitionable Devices", Label("operator", "dra-pd"), Ordered, func() {
+	var (
+		initialKueueInstance *ssv1.Kueue
+		migMemoryBytes       int64  // 1g.10gb memory in bytes
+		migProfile           string // "1g.10gb"
+		secondMIGMemoryBytes int64  // 3g.20gb or 3g.40gb memory in bytes
+		secondMIGProfile     string // "3g.20gb" or "3g.40gb"
+		migDeviceCount       int
+		secondMIGDeviceCount int
+		wholeGPUMemoryBytes  int64
+		hasWholeGPUDevice    bool
+	)
+
+	JustAfterEach(func(ctx context.Context) {
+		testutils.DumpKueueControllerManagerLogs(ctx, kubeClient, 500)
+	})
+
+	BeforeAll(func(ctx context.Context) {
+		if !testutils.IsDRASupported(kubeClient) {
+			Skip("DRA APIs (resource.k8s.io/v1) not available on this cluster")
+		}
+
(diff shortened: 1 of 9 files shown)


sohankunkerkar:
You need to add `gpu-late-dc` to https://github.com/openshift/kueue-operator/blob/main/test/e2e/bindata/assets/08_kueue_default.yaml#L15

This might address your CI issue

pkg/operator/target_config_reconciler.go
```diff
@@ -97,29 +97,30 @@ const (
...
+	operatorNamespace              string
+	resourceCache                  resourceapply.ResourceCache
+	kueueImage                     string
+	serviceMonitorSupport          bool
+	apiRegistrationClient          apiregistrationv1client.ApiregistrationV1Interface
+	openshiftConfigClient          configclient.Interface
+	configInformer                 dynamicinformer.DynamicSharedInformerFactory
+	isOpenShift                    bool
+	draExtendedResourceEnabled     bool
+	draPartitionableDevicesEnabled bool
```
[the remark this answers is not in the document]
PannagaRao: But preserving previous state was added intentionally. If the API server is temporarily unavailable, we don't want to flip the gates to false and cause kueue to restart without DRA support. The current value is likely still correct.

test/e2e/e2e_dra_partitionable_devices_test.go
```diff
@@ -0,0 +1,934 @@
...
+					g.Expect(res.Total.Value()).To(Equal(expectedBytes))
+					return
+				}
+			}
+		}
+		g.Expect(false).To(BeTrue(), "gpu.memory not found in CQ status")
+	}, testutils.OperatorReadyTime, testutils.OperatorPoll).Should(Succeed())
+}
+
+var _ = Describe("DRA Partitionable Devices", Label("operator", "dra-pd"), Ordered, func() {
```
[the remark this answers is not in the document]
PannagaRao: PD tests need a MIG-configured cluster (A100/H100) — they can't run on the standard DRA CI cluster which uses T4 GPUs with no MIG. So these tests won't be run regularly like the others

test/e2e/e2e_dra_partitionable_devices_test.go
```diff
@@ -0,0 +1,934 @@
...
+		if fg.Spec.FeatureSet == configv1.CustomNoUpgrade && fg.Spec.CustomNoUpgrade != nil {
+			for _, gate := range fg.Spec.CustomNoUpgrade.Enabled {
+				if string(gate) == "DRAPartitionableDevices" {
+					pdGateEnabled = true
+					break
+				}
+			}
+		}
+		if !pdGateEnabled {
+			Skip("DRAPartitionableDevices K8s feature gate not enabled via CustomNoUpgrade")
```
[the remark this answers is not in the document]
PannagaRao: Currently DRAPartitionableDevices is alpha and can be set only by CustomNoUpgrade

test/e2e/e2e_dra_partitionable_devices_test.go
```diff
@@ -0,0 +1,934 @@
...
+							secondMIGMemoryBytes = mem
+							secondMIGProfile = "3g.40gb"
+						}
+						if secondMIGProfile == "3g.40gb" {
+							secondMIGDeviceCount++
+						}
+					}
+				}
+			}
+		}
```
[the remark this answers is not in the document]
PannagaRao: Mixed GPU clusters (e.g., A100 + H100 on the same node) are not a realistic test environment for us.

test/e2e/e2e_dra_partitionable_devices_test.go
```diff
@@ -0,0 +1,934 @@
...
+							DeviceSelector: ssv1.DeviceSelector{
+								Type: ssv1.DeviceSelectorTypeCEL,
+								CEL:  ssv1.CELDeviceSelector{Expression: "device.driver == 'gpu.nvidia.com'"},
+							},
+						},
+					},
+				},
+			},
+		}
+		applyKueueConfig(ctx, kueueInstance.Spec.Config, kubeClient)
```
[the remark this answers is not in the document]
PannagaRao: This is valid. Will add a check after applyKueueConfig to verify KueueDRAIntegrationPartitionableDevices is present in the config

test/e2e/e2e_dra_partitionable_devices_test.go
```diff
@@ -0,0 +1,934 @@
...
+		builder := testutils.NewTestResourceBuilder(ns.Name, pdLocalQueueName)
+		pod := builder.NewPod()
+		pod.Spec.Containers[0].Resources.Requests[corev1.ResourceCPU] = resource.MustParse("100m")
+		pod.Spec.Containers[0].Resources.Requests[corev1.ResourceMemory] = resource.MustParse("100Mi")
+		addDRAClaims(&pod.Spec, "mig-pod")
+		createdPod, err := kubeClient.CoreV1().Pods(ns.Name).Create(ctx, pod, metav1.CreateOptions{})
+		Expect(err).NotTo(HaveOccurred())
+		DeferCleanup(func() {
+			_ = kubeClient.CoreV1().Pods(ns.Name).Delete(context.TODO(), createdPod.Name, metav1.DeleteOptions{})
+		})
```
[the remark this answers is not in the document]
PannagaRao: This is the existing pattern in the repo

upstream/kueue/src
```diff
@@ -1 +1 @@
-Subproject commit 851a097d19c4326bdfdf0276492625e17a08495e
+Subproject commit 982b8b029a7f5215c3bf3b2b426f6cfd765a8ee2
```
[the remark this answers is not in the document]
PannagaRao: No longer part of this PR

PannagaRao:
This PR is ready for review.

/assign @sohankunkerkar @kannon92

test/e2e/e2e_dra_partitionable_devices_test.go
```diff
@@ -0,0 +1,946 @@
...
+		wholeGPUMemoryBytes  int64
+		hasWholeGPUDevice    bool
+	)
+
+	JustAfterEach(func(ctx context.Context) {
+		testutils.DumpKueueControllerManagerLogs(ctx, kubeClient, 500)
+	})
+
+	BeforeAll(func(ctx context.Context) {
+		draSupported := false
```
kannon92: can we refactor this into a function and reuse this for our DRA testing?

can be done as a followup.

test/e2e/e2e_dra_partitionable_devices_test.go
```diff
@@ -0,0 +1,946 @@
...
+		}, testutils.OperatorReadyTime, testutils.OperatorPoll).Should(Succeed())
+	})
+
+	It("should charge largest counter value when CEL matches multiple device types", func(ctx context.Context) {
+		if secondMIGMemoryBytes == 0 {
+			Skip("No second MIG profile found")
+		}
+		kueueClient := clients.UpstreamKueueClient
+
+		By("Creating ResourceFlavor, ClusterQueue, Namespace and LocalQueue")
```
kannon92: CAn you refactor all this setup to be a function that you call with just the few changes?

there is a lot of duplicated setup making this harder to review.

test/e2e/testutils/builders.go
```diff
@@ -360,3 +361,47 @@ func (b *TestResourceBuilder) NewLeaderWorkerSet(opts LeaderWorkerSetOptions) *l
...
+	}
+	if celExpression != "" {
+		rct.Spec.Spec.Devices.Requests[0].Exactly.Selectors = []resourcev1.DeviceSelector{
+			{CEL: &resourcev1.CELDeviceSelector{Expression: celExpression}},
+		}
+	}
+	return rct
+}
+
+func (b *TestResourceBuilder) NewDRAJob(name, queueName, templateName string) *batchv1.Job {
```
anahas-redhat, line 415: The new NewDRAJob does essentially the same thing as the existing newDRAJob in [e2e_dra_test.go](https://github.com/openshift/kueue-operator/blob/main/test/e2e/e2e_dra_test.go#L1415). By that I mean, both create a job with a DRA claim, CPU/memory requests, and a sleep command. The difference is that the existing one uses setDRAJobCPU and sleep 10, while the new one inlines 100m/100Mi and sleep 120. Could we consolidate into one, or have the new one extend the existing one?

test/e2e/testutils/builders.go
```diff
@@ -360,3 +361,47 @@ func (b *TestResourceBuilder) NewLeaderWorkerSet(opts LeaderWorkerSetOptions) *l
 
 	return lws
 }
+
+func NewResourceClaimTemplate(name, namespace, deviceClass string, count int64, celExpression string) *resourcev1.ResourceClaimTemplate {
```
anahas-redhat, line 369: The new NewResourceClaimTemplate does essentially the same thing as the existing newDRAResourceClaimTemplate in[ e2e_dra_test.go](https://github.com/openshift/kueue-operator/blob/main/test/e2e/e2e_dra_test.go#L1324) but adds deviceClass and celExpression parameters, which makes it more flexible. Could we consolidate into one? That way we avoid duplication.

test/e2e/e2e_dra_partitionable_devices_test.go
```diff
@@ -0,0 +1,946 @@
...
+							DeviceSelector: ssv1.DeviceSelector{
+								Type: ssv1.DeviceSelectorTypeCEL,
+								CEL:  ssv1.CELDeviceSelector{Expression: "device.driver == 'gpu.nvidia.com'"},
+							},
+						},
+					},
+				},
+			},
+		}
+		applyKueueConfig(ctx, kueueInstance.Spec.Config, kubeClient)
```
anahas-redhat, line 469: Worth adding the check after applyKueueConfig to verify KueueDRAIntegrationPartitionableDevices, as CodeRabbit mentioned.

test/e2e/e2e_dra_partitionable_devices_test.go
```diff
@@ -0,0 +1,946 @@
...
+		By("Submitting first job with sources configured")
+		job1 := builder.NewDRAJob("pd-runtime-1", pdLocalQueueName, "mig-runtime")
+		job1.Spec.Template.Spec.Containers[0].Command = []string{"sh", "-c", "sleep 600"}
+		createdJob1, err := kubeClient.BatchV1().Jobs(ns.Name).Create(ctx, job1, metav1.CreateOptions{})
+		Expect(err).NotTo(HaveOccurred())
+		DeferCleanup(testutils.CleanUpJob, kubeClient, ns.Name, createdJob1.Name)
+
+		verifyWorkloadGPUMemory(ctx, ns.Name, string(createdJob1.UID), migMemoryBytes)
+
+		By("Removing sources from Kueue CR")
```
anahas-redhat, line 704: The "should fall back to device-count" test removes sources mid-test but restores them before the test ends (line ~574). This test also removes but doesn't restore them — it relies on AfterAll to clean up. Could we add a DeferCleanup or an explicit restore at the end, like the fallback test does? That way if someone adds a new It after this one, it won't silently run without counter sources.

test/e2e/e2e_dra_partitionable_devices_test.go
PannagaRao: The check after restoring sources is already present — line ~584 verifies KueueDRAIntegrationPartitionableDevices is back in the configmap before the next test runs.

test/e2e/e2e_dra_partitionable_devices_test.go
PannagaRao: The alpha limitation test intentionally relies on AfterAll for cleanup. Adding an explicit restore here would cause a double restore — AfterAll would then try to restore a config that's already been restored, and it fails because the deployment resourceVersion doesn't change between the two restores. This test is ordered last for this reason. If the next implementation phase handles the counter/device-count transition cleanly, this limitation test will be obsolete.

test/e2e/e2e_dra_partitionable_devices_test.go
PannagaRao: Refactored into a helper that determines if DRA is supported. Will update other dra test files to use it as a follow-up PR

test/e2e/e2e_dra_partitionable_devices_test.go
PannagaRao: Done. PTAL!

test/e2e/testutils/builders.go
PannagaRao: Will consolidate these in a follow-up PR that refactors the DRA test suites to use the shared helpers. Keeping this PR minimal for review.

test/e2e/testutils/builders.go
PannagaRao: The existing one will be replaced by the new one added in builders.go as part of follow-up PR that refactors the DRA test. Keeping this PR minimal for review.

test/e2e/testutils/builders.go
anahas-redhat: I'm fine deferring the full refactor, but could you at least have the new functions call the existing ones (or vice versa) so we don't merge two independent implementations of the same thing? The follow-up then becomes a rename/move, not a dedup.

test/e2e/testutils/builders.go
anahas-redhat: I'm fine deferring the full refactor, but could you at least have the new functions call the existing ones (or vice versa) so we don't merge two independent implementations of the same thing? The follow-up then becomes a rename/move, not a dedup.

test/e2e/testutils/builders.go
PannagaRao: Handled in https://github.com/openshift/kueue-operator/pull/2031

test/e2e/testutils/builders.go
PannagaRao: Handled in https://github.com/openshift/kueue-operator/pull/2031

kannon92 (commented):
/lgtm
/cherry-pick release-1.4

PR #1991, opened 2026-06-09
Add default usage half life time

MaysaMacedo:
## Summary by CodeRabbit

* **New Features**
  * Admission fair sharing adds a required mode selector (Default | Custom); Custom enables nested per-field tuning.

* **Chores**
  * Schema/CRD updated to require the mode and conditionally allow nested custom settings.
  * Operator config generation updated to respect the selected mode.
  * Build tooling updated to use a newer Go toolchain for client generation.

* **Tests**
  * Unit, envtest, and e2e tests extended to cover the new configuration shape and validations.


pkg/apis/kueueoperator/v1/types.go
```diff
@@ -135,9 +135,36 @@ const (
...
+	// Custom means admission fair sharing is enabled with user-provided configuration
+	// in the custom field.
+	// +required
+	// +unionDiscriminator
+	Configuration AdmissionFairSharingConfiguration `json:"configuration,omitempty"`
+	// custom provides customized configuration for admission fair sharing.
+	// custom is required when configuration is Custom, and forbidden otherwise.
+	// +optional
+	Custom AdmissionFairSharingCustom `json:"custom,omitzero"`
+}
```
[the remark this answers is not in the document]
MaysaMacedo: The previous version of This field is not yet in use by any user. So, there is no need to worry about compatibility window.

pkg/configmap/configmap_test.go
```diff
@@ -1057,16 +1057,77 @@ webhook:
...
+  enableClusterQueueResources: true
+namespace: test
+webhook:
+  port: 9443
+`,
+				},
+			},
+			wantErr: nil,
+		},
+		"Admission Fair Sharing with usageHalfLifeTime, usageSamplingInterval and resourceWeights set": {
```
kannon92, line 1117: Can we add a test case where resourceWeights fails to parse?

MaysaMacedo: @kannon92 That is handled in the cel tests. Let me know if you would like any additional test there.

pkg/configmap/configmap.go
```diff
@@ -272,24 +272,30 @@ func buildFeatureGates(frameworks []kueue.KueueIntegration, draExtendedResourceE
...
+			result.UsageHalfLifeTime = v1.Duration{Duration: time.Duration(custom.UsageHalfLifeTimeSeconds) * time.Second}
+		}
+		if custom.UsageSamplingIntervalSeconds > 0 {
+			result.UsageSamplingInterval = v1.Duration{Duration: time.Duration(custom.UsageSamplingIntervalSeconds) * time.Second}
+		}
+		resourceWeights := make(map[corev1.ResourceName]float64, len(custom.ResourceWeights))
+		for _, entry := range custom.ResourceWeights {
+			f, err := strconv.ParseFloat(entry.Weight, 64)
+			if err != nil {
+				return nil, fmt.Errorf("failed to parse resource weight for resource %s: %w", entry.Name, err)
```
kannon92, line 291: Is this code reachable?

Seems we do this validation in CEL.

pkg/configmap/configmap_test.go
kannon92: https://github.com/openshift/kueue-operator/pull/1991/changes#r3389662542

I just don't think we cover the error test case at all right now.

kannon92 (commented):
/hold

/lgtm

Should have an api-review but this aligns with what I expect.

pkg/configmap/configmap.go
MaysaMacedo: I believe it's worth to keep it here in case there is any corner case that is not already covered in CEL, instead of silently ignoring the error.

kannon92 (commented):
/lgtm

MaysaMacedo:
e2e tests passed. The failure is from must-gather, unrelated to the PR.

pkg/apis/kueueoperator/v1/types.go
```diff
@@ -135,9 +135,36 @@ const (
...
+)
+
 // AdmissionFairSharing provides configuration of intervals
 // and resource weights for Admission Fair Sharing.
+// +kubebuilder:validation:XValidation:rule="has(self.configuration) && self.configuration == 'Custom' ? has(self.custom) : !has(self.custom)",message="custom is required when configuration is Custom, and forbidden otherwise"
+// +union
 type AdmissionFairSharing struct {
+	// configuration determines if admission fair sharing uses default or custom settings.
+	// The allowed values are Default and Custom.
+	// Default means admission fair sharing is enabled with kueue's defaults,
```
everettraven: These aren't actually kueue's defaults anymore right? We may want to be more clear that this is our opinionated defaults as opposed to the defaults of kueue itself.

everettraven (commented):
Aside from a minor comment on clarifying the defaulting language a bit more to prevent confusion, the API change LGTM.

pkg/apis/kueueoperator/v1/types.go
MaysaMacedo: It's both. For UsageHalfLifeTimeSeconds it's the kueue-operator default, while for UsageSamplingIntervalSeconds and ResourceWeights is kueue defaults. But I can adjust the wording.

everettraven (approved):
LGTM

MaysaMacedo:
I had to regenerate the bundle to include the docs changes.

PR #1996, opened 2026-06-11
Add gpu-late-dc DeviceClass mapping for upstream DRA e2e tests

PannagaRao:
The upstream baseline DRA test "Should admit job after DeviceClass is created" creates a DeviceClass gpu-late-dc.example.com at runtime and expects kueue to translate it to the logical name gpu-late-dc. Without this mapping in the default Kueue CR, the workload is never admitted on OpenShift because the operator controls the kueue config.


PannagaRao:
We need this fix for the upstream dra e2e
cc: @sohankunkerkar @kannon92

PR #2000, opened 2026-06-11
Add default afs scenario

MaysaMacedo:
Assisted-by: Claude

## Summary by CodeRabbit

* **Tests**
  * Expanded end-to-end coverage for Admission Fair Sharing across default and custom configurations.
  * Added an ordered scenario that verifies cluster saturation behavior: newly created jobs remain suspended while the cluster queue is full, then one becomes admitted after capacity is freed.
  * Strengthened test assertions for usage half-life, sampling interval, and custom resource weight configuration behavior.

### DIFF
--- a/test/e2e/e2e_admission_fair_sharing_test.go
+++ b/test/e2e/e2e_admission_fair_sharing_test.go
@@ -46,7 +46,13 @@ var _ = Describe("Admission Fair Sharing", Label("admission-fair-sharing"), Orde
 	})
 
 	BeforeAll(func(ctx context.Context) {
-		enableAdmissionFairSharing(ctx, usageHalfLifeTimeSeconds, usageSamplingIntervalSeconds)
+		enableAdmissionFairSharing(ctx, ssv1.AdmissionFairSharing{
+			Configuration: ssv1.AdmissionFairSharingConfigurationCustom,
+			Custom: ssv1.AdmissionFairSharingCustom{
+				UsageHalfLifeTimeSeconds:     usageHalfLifeTimeSeconds,
+				UsageSamplingIntervalSeconds: usageSamplingIntervalSeconds,
+			},
+		})
 	})
 
 	When("LocalQueues have different FairSharing weight values", func() {
@@ -428,10 +434,17 @@ var _ = Describe("Admission Fair Sharing", Label("admission-fair-sharing"), Orde
 
 		It("should deprioritize the queue with higher weighted CPU usage when cpu weight is 10 and memory weight is 0", func(ctx context.Context) {
 			By("Overriding Kueue config with resourceWeights (cpu: 10, memory: 0) and fast decay")
-			enableAdmissionFairSharing(ctx, 10, 1,
-				ssv1.ResourceWeight{Name: "cpu", Weight: "10"},
-				ssv1.ResourceWeight{Name: "memory", Weight: "0"},
-			)
+			enableAdmissionFairSharing(ctx, ssv1.AdmissionFairSharing{
+				Configuration: ssv1.AdmissionFairSharingConfigurationCustom,
+				Custom: ssv1.AdmissionFairSharingCustom{
+					UsageHalfLifeTimeSeconds:     10,
+					UsageSamplingIntervalSeconds: 1,
+					ResourceWeights: []ssv1.ResourceWeight{
+						{Name: "cpu", Weight: "10"},
+						{Name: "memory", Weight: "0"},
+					},
+				},
+			})
 
 			By("Creating Resource Flavor")
 			resourceFlavor, cleanupResourceFlavor, err := testutils.NewResourceFlavor().WithGenerateName().CreateWithObject(ctx, clients.UpstreamKueueClient)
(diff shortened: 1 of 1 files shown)


test/e2e/e2e_admission_fair_sharing_test.go
```diff
@@ -579,25 +721,25 @@ func enableAdmissionFairSharing(ctx context.Context, halfLifeSeconds, samplingSe
...
+		case ssv1.AdmissionFairSharingConfigurationCustom:
+			expectedHalfLife := (time.Duration(afs.Custom.UsageHalfLifeTimeSeconds) * time.Second).String()
+			g.Expect(configData).To(ContainSubstring(fmt.Sprintf("usageHalfLifeTime: %s", expectedHalfLife)), "usageHalfLifeTime not found in ConfigMap")
+		}
+		expectedSampling := (time.Duration(afs.Custom.UsageSamplingIntervalSeconds) * time.Second).String()
 		g.Expect(configData).To(ContainSubstring(fmt.Sprintf("usageSamplingInterval: %s", expectedSampling)), "usageSamplingInterval not found in ConfigMap")
-		for _, rw := range resourceWeights {
+		for _, rw := range afs.Custom.ResourceWeights {
 			g.Expect(configData).To(ContainSubstring(fmt.Sprintf("%s: %s", rw.Name, rw.Weight)),
 				fmt.Sprintf("resourceWeight %s: %s not found in ConfigMap", rw.Name, rw.Weight))
```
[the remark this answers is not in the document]
MaysaMacedo: I don't understand the problem.

test/e2e/e2e_admission_fair_sharing_test.go
```diff
@@ -564,9 +577,138 @@ var _ = Describe("Admission Fair Sharing", Label("admission-fair-sharing"), Orde
...
+				PropagationPolicy: ptr.To(metav1.DeletePropagationBackground),
+			})
+			Expect(err).NotTo(HaveOccurred(), "Failed to delete job1")
+
+			By("Verifying job3 is admitted")
+			Eventually(func(g Gomega) {
+				g.Expect(testutils.IsJobSuspended(ctx, kubeClient, ns.Name, job3.Name)).To(BeFalse())
+			}, testutils.OperatorReadyTime, testutils.OperatorPoll).Should(Succeed(),
+				"job %s was not admitted", job3.Name)
+		})
```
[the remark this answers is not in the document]
MaysaMacedo: waiting for localqueue consumed resources to be filled would increase a lot the time the time of the test execution, up to 5 minutes. So, I decided to remove that check.

test/e2e/e2e_admission_fair_sharing_test.go
```diff
@@ -682,9 +693,134 @@ var _ = Describe("Admission Fair Sharing", Label("admission-fair-sharing"), Orde
...
+			Expect(err).NotTo(HaveOccurred())
+
+			By("Creating LocalQueue lq-b bound to the ClusterQueue")
+			lqB, cleanupLQB, err = testutils.NewLocalQueue(ns.Name, "lq-b").
+				WithClusterQueue(cq.Name).
+				CreateWithObject(ctx, clients.UpstreamKueueClient)
+			Expect(err).NotTo(HaveOccurred())
+		})
+
+		AfterEach(func(ctx context.Context) {
```
anahas-redhat: Once for now we have only one It, can we delete the resource created with DeferCleanUp so we can remove the AfterEach? 
For example: DeferCleanup(cleanupResourceFlavor), DeferCleanup(cleanupClusterQueue), DeferCleanup(cleanupNs) etc?

test/e2e/e2e_admission_fair_sharing_test.go
```diff
@@ -682,9 +693,134 @@ var _ = Describe("Admission Fair Sharing", Label("admission-fair-sharing"), Orde
...
+		BeforeAll(func(ctx context.Context) {
+
+			By("Enabling AdmissionFairSharing with default values")
+			enableAdmissionFairSharing(ctx, ssv1.AdmissionFairSharing{
+				Configuration: ssv1.AdmissionFairSharingConfigurationDefault,
+			})
+
+		})
+
+		BeforeEach(func(ctx context.Context) {
```
anahas-redhat: Since there's only one It in this block, could we move the resource creation (RF, CQ, namespace, LQs) into the It itself with DeferCleanup, and drop the BeforeEach/AfterEach? This would match the pattern used in the rest of this file (e.g., the custom AFS  tests above) and in e2e_dra_test.go, where BeforeAll handles the cluster-level config and each It is self-contained.

test/e2e/e2e_admission_fair_sharing_test.go
```diff
@@ -697,27 +833,28 @@ func enableAdmissionFairSharing(ctx context.Context, halfLifeSeconds, samplingSe
...
+		case ssv1.AdmissionFairSharingConfigurationDefault:
+			// The operator uses a 30-minute default for usageHalfLifeTime.
+			g.Expect(configData).To(ContainSubstring("usageHalfLifeTime: 30m0s"), "usageHalfLifeTime not found in ConfigMap")
+			g.Expect(configData).To(ContainSubstring("usageSamplingInterval: 0s"), "usageSamplingInterval not found in ConfigMap")
+		case ssv1.AdmissionFairSharingConfigurationCustom:
+			expectedHalfLife := (time.Duration(afs.Custom.UsageHalfLifeTimeSeconds) * time.Second).String()
+			g.Expect(configData).To(ContainSubstring(fmt.Sprintf("usageHalfLifeTime: %s", expectedHalfLife)), "usageHalfLifeTime not found in ConfigMap")
+			expectedSampling := (time.Duration(afs.Custom.UsageSamplingIntervalSeconds) * time.Second).String()
+			g.Expect(configData).To(ContainSubstring(fmt.Sprintf("usageSamplingInterval: %s", expectedSampling)), "usageSamplingInterval not found in ConfigMap")
+			for _, rw := range afs.Custom.ResourceWeights {
```
anahas-redhat, line 830: We're only checking resourceWeights in case of Custom, right? In case of Default, can we add something like NotContainSubstring("resourceWeights")? Just to make sure it's not set (if that's the case).

anahas-redhat (commented):
Thanks for creating this test case. I've added my comments below.

anahas-redhat:
Thanks for applying the changes Maysa.
/lgtm
/approve

PR #2009, opened 2026-06-11
Fix API lint: add omitempty to union discriminator Type fields

PannagaRao:

## Bug Fixes
* API responses now exclude unset discriminator fields from JSON output, resulting in cleaner and more compact responses.


PannagaRao:
cc: @MaysaMacedo @kannon92

PR #2108, opened 2026-06-29
Fix e2e upgrade test failures caused by operator resync timing

sohankunkerkar:
The operator's library-go factory controller uses ResyncEvery(5m). When a test updates the Kueue CR, the informer event may be coalesced with an in-flight sync that reads stale CR data from the cache. The next sync that picks up the new CR data doesn't happen until the periodic resync, which is 5 minutes away.

Three fixes:
- applyKueueConfig: replace deployment-ResourceVersion-based wait with a ConfigMap content-based wait. Instead of checking if the deployment version changed (can be triggered by unrelated reconciliation), verify the ConfigMap YAML actually differs from its pre-update content.
- TLS BeforeAll: add Eventually poll for TLS options to appear in the ConfigMap. After an upgrade, the operator's initial sync may be delayed by certificate readiness waits.
- OperatorReadyTime: increase from 3m to 6m to cover the 5m resync period as a safety net.

xref: https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_release/80689/rehearse-80689-pull-ci-openshift-kueue-operator-release-1.4-upgrade-from-4.22-e2e-operator-upgrade-4-22-kueue-1-3-to-1-4/2070524808960937984

## Summary by CodeRabbit

* **Tests**
  * Improved end-to-end checks to wait for the operator-updated configuration to propagate before validating operator and TLS-related behavior, reducing timing-related flakiness.
  * Increased the operator readiness wait time used by tests to improve stability on slower environments.

### DIFF
--- a/test/e2e/e2e_operator_test.go
+++ b/test/e2e/e2e_operator_test.go
@@ -21,7 +21,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
-	"reflect"
+
 	"slices"
 	"strings"
 	"time"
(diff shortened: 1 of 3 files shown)


anahas-redhat:
It seems more valid to wait for kueue-manager-config content than kueue-controller-manager ResourceVersion — the ConfigMap is the actual artifact the controller consumes, while the deployment RV can change for unrelated reasons. 
TLS implementation is more robust now too with the Eventually poll. 
Increasing OperatorReadyTime to 6m seems inevitable given the operator's ResyncEvery takes at least 5 minutes — the old 3m timeout couldn't survive the worst-case resync delay. 

/lgtm
/approve

PR #2140, opened 2026-07-03
Alow configuration of waitForPodsReady feature

MaysaMacedo:
These are the fields that are being enabled in this PR https://github.com/kubernetes-sigs/kueue/blob/23205b887b314a3faf6f3ea9ac2913629708aefc/apis/config/v1beta2/configuration_types.go#L304-L329

Note that `GangScheduling` is the same as `WaitForPodsReady`.

This PR was developed with assistance of Claude.

## Summary by CodeRabbit

* **New Features**
  * Added configurable gang-scheduling admission and recovery timeouts.
  * Added requeuing controls, including retry limits, exponential backoff, and eviction or creation time references.

* **Behavior Changes**
  * Removed the unsupported `None` gang-scheduling policy option.
  * Updated default admission behavior and configuration descriptions.

* **Tests**
  * Expanded coverage for timeout, recovery, blocking, and requeuing configurations.


MaysaMacedo:
/cc @kannon92
when you can, could you provide some input into the TBD stated in the PR description? Thanks

pkg/apis/kueueoperator/v1/types.go
```diff
@@ -365,6 +365,64 @@ type ByWorkload struct {
...
+	// backoffBaseSeconds defines the base for the exponential backoff for re-queuing an evicted workload.
+	// The consecutive requeue delays are computed as backoffBaseSeconds * 2^(retryCount-1) + jitter.
+	// Defaults to 60.
+	// +optional
+	BackoffBaseSeconds *int32 `json:"backoffBaseSeconds,omitempty"`
+
+	// backoffMaxSeconds defines the maximum backoff time to re-queue an evicted workload.
+	// Defaults to 3600.
+	// +optional
+	BackoffMaxSeconds *int32 `json:"backoffMaxSeconds,omitempty"`
```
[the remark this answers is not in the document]
kannon92: Good bot.

pkg/apis/kueueoperator/v1/types.go
```diff
@@ -365,6 +363,73 @@ type ByWorkload struct {
...
 	// +required
 	Admission GangSchedulingWorkloadAdmission `json:"admission"`
+
+	// timeoutSeconds defines the time for an admitted workload to have it's pods scheduled.
+	// When the timeoutSeconds is expired, the workload is evicted and requeued.
+	// The value must be between 1 and 31536000 (one year in seconds). Defaults to 1800 seconds.
+	// +kubebuilder:validation:Minimum=1
+	// +kubebuilder:validation:Maximum=31536000
+	// +optional
+	TimeoutSeconds int32 `json:"timeoutSeconds,omitempty"`
```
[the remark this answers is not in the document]
MaysaMacedo: When the timeout Duration is 0s the kueue-controller-manager internally sets it to 30 minutes
https://github.com/kubernetes-sigs/kueue/blob/main/apis/config/v1beta2/defaults.go#L104-L108

bundle/manifests/kueue.openshift.io_kueues.yaml
```diff
@@ -179,24 +179,93 @@ spec:
...
-                          None means gang scheduling is disabled.
                           When set to "", this means no opinion and the operator is left
                           to choose a reasonable default, which is subject to change over time.
-                          The current default is None.
+                          The current default is "".
                           policy is a required field.
                         enum:
                         - ByWorkload
-                        - None
                         - ""
```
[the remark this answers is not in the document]
MaysaMacedo: This was removed to accommodate the fact that the feature is always enabled by default https://github.com/kubernetes-sigs/kueue/issues/11366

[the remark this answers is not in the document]
MaysaMacedo: @everettraven We have discussed about this topic before. Do you consider the issue detailed here in this thread a valid concern?

MaysaMacedo:
@kannon92 @sohankunkerkar @PannagaRao
Could you take a look at this PR? thank you!

pkg/apis/kueueoperator/v1/types.go
```diff
@@ -308,12 +308,11 @@ type LabelKeys struct {
 	Key string `json:"key,omitempty"`
 }
 
-// +kubebuilder:validation:Enum=ByWorkload;None;""
+// +kubebuilder:validation:Enum=ByWorkload;""
 type GangSchedulingPolicy string
 
 const (
 	GangSchedulingPolicyByWorkload GangSchedulingPolicy = "ByWorkload"
```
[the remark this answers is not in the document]
kannon92: Removing an enum is considered a breaking change. Why was this removed?

pkg/apis/kueueoperator/v1/types.go
```diff
@@ -308,12 +308,11 @@ type LabelKeys struct {
 	Key string `json:"key,omitempty"`
 }
 
-// +kubebuilder:validation:Enum=ByWorkload;None;""
+// +kubebuilder:validation:Enum=ByWorkload;""
```
kannon92: Removing an enum is considered a breaking change. Why did we remove None?

kannon92: Oh yea because it None was removed from kueue..

kannon92:
/hold

Could we get @everettraven to take a look?

pkg/apis/kueueoperator/v1/types.go
MaysaMacedo: yeah, our API in the Kueue-operator mentions:
- `None` means gang scheduling is disabled.
- while `""`  means no opinion and the operator is left to choose a reasonable default.
Given the kueue upstream always enable Kueue, I removed the option of defining the `None`.
Now, I wanted to hear opinion of others if we should really follow this path...

bundle/manifests/kueue.openshift.io_kueues.yaml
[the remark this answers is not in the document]
kannon92: IMO this concern is legit. What would happen to users configurations if they specified None before and then upgrade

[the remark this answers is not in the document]
everettraven: My understanding was that the ability to disable the functionality was being removed entirely as part of the change to enable the functionality by default (in reality this is forcing to always enabled with some default behaviors).

What is the minimum version of OpenShift that this change going to be compatible with? If upgrades to this version of the API are limited to versions of OpenShift based on k8s 1.33+, this should fall under ratcheting validation behaviors and allow spec updates as long as the field with the changed validation is not modified.

For more information on ratcheting validation, here are the docs: https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#validation-ratcheting

Before we merge this, can we validate that this does indeed ratchet? How we test that is:
1. Install the previous iteration of the CRD
2. Create a resource using the old allowed value
3. Install the new iteration of the CRD
4. Ensure we can update any other field without having to modify the field with the validation change
5. Ensure we can update the field with the validation change to a new allowed value
6. Ensure we cannot update the field from a new allowed value to the now removed value.

We have automated the testing for this in openshift/api but I'd be OK with a manual verification of this.

pkg/apis/kueueoperator/v1/types.go
```diff
@@ -308,12 +308,11 @@ type LabelKeys struct {
...
 type GangSchedulingPolicy string
 
 const (
 	GangSchedulingPolicyByWorkload GangSchedulingPolicy = "ByWorkload"
-	GangSchedulingPolicyNone       GangSchedulingPolicy = "None"
```
everettraven, line 316: I would not remove this yet - even though you've removed the allowed value, your controllers must still understand what to do if they see this value until you can be reasonably confident that your users will no longer have this value set (in this case, this will likely stick around for a while). You can also never add this back as an allowed value for the field because it would end up breaking clients.

You may want to leave a comment explaining this for future reference.

pkg/apis/kueueoperator/v1/types.go
```diff
@@ -329,26 +328,25 @@ const (
...
 	// to choose a reasonable default, which is subject to change over time.
-	// The current default is None.
+	// The current default is "".
 	// policy is a required field.
 	// +required
 	// +unionDiscriminator
 	Policy GangSchedulingPolicy `json:"policy"`
-	// byWorkload configures how Kueue will process workloads for admission.
+	// byWorkload configures how Kueue will handle workloads that are not ready within a timeout.
 	// byWorkload is required when policy is ByWorkload, and forbidden otherwise.
```
everettraven, line 349: `ByWorkload` is equivalent to `WaitForPodsReady` in Kueue?

If so, would something like:
> byWorkload configures the workload readiness evaluation criteria that Kueue uses for workload scheduling.
>...

be more accurate?

pkg/apis/kueueoperator/v1/types.go
```diff
@@ -365,6 +363,82 @@ type ByWorkload struct {
 	// The current default is Parallel.
 	// +required
 	Admission GangSchedulingWorkloadAdmission `json:"admission"`
```
everettraven, line 370: Just to verify, setting this field to `Sequential` is equivalent to configuring [Kueue's `BlockAdmission` field](https://github.com/kubernetes-sigs/kueue/blob/23205b887b314a3faf6f3ea9ac2913629708aefc/apis/config/v1beta2/configuration_types.go#L310-L314) as `true`?

pkg/apis/kueueoperator/v1/types.go
```diff
@@ -365,6 +363,82 @@ type ByWorkload struct {
...
+
+	// timeoutSeconds defines the time for an admitted workload to reach condition PodsReady=true.
+	// When the timeoutSeconds is expired, the workload is evicted and requeued.
+	// The value must be between 1 and 31536000 (one year in seconds).
+	// When omitted, kueue will decide the default, which is subject to change over time.
```
everettraven: For future-proofing, I'd suggest that we modify this to say:

```suggestion
	// When omitted, this means no opinion and the operator is left to choose a reasonable default, which is subject to change over time.
```

This allows us to modify the default behavior as needed for OpenShift without being bound to the whims of whatever upstream Kueue decides to do. Our default behavior can match that of upstream Kueue still with this approach.

pkg/apis/kueueoperator/v1/types.go
```diff
@@ -365,6 +363,82 @@ type ByWorkload struct {
...
+	// The current default is 1800 seconds.
+	// +kubebuilder:validation:Minimum=1
+	// +kubebuilder:validation:Maximum=31536000
```
everettraven: How was this minimum and maximum decided?

pkg/apis/kueueoperator/v1/types.go
everettraven: This applies to everywhere we are doing defaulting.

bundle/manifests/kueue.openshift.io_kueues.yaml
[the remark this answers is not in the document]
MaysaMacedo: @everettraven It would need to be compatible with OCP 4.18+.

pkg/apis/kueueoperator/v1/types.go
```diff
@@ -365,6 +363,82 @@ type ByWorkload struct {
...
+	// CreationTimestamp requeues the workload based on its .metadata.creationTimestamp.
+	CreationTimestamp RequeuingTimeReference = "Creation"
+)
+
+// RequeuingStrategy defines the strategy for requeuing a Workload evicted.
+type RequeuingStrategy struct {
+	// timeReference defines which point in time is used as the reference for re-queuing a Workload
+	// that was evicted. Allowed values are Eviction and Creation.
+	// Eviction uses the time of the Evicted condition with PodsReadyTimeout reason.
+	// Creation uses the Workload .metadata.creationTimestamp.
```
everettraven: Not a huge deal because this matches the upstream, but something like `EvictionTime` and `CreationTime` reads as more descriptive values to me.

pkg/apis/kueueoperator/v1/types.go
```diff
@@ -365,6 +363,82 @@ type ByWorkload struct {
...
+
+const (
+	// EvictionTimestamp requeues the workload based on the time of the Evicted condition
+	// with PodsReadyTimeout reason.
+	EvictionTimestamp RequeuingTimeReference = "Eviction"
+	// CreationTimestamp requeues the workload based on its .metadata.creationTimestamp.
+	CreationTimestamp RequeuingTimeReference = "Creation"
+)
+
+// RequeuingStrategy defines the strategy for requeuing a Workload evicted.
```
everettraven:
```suggestion
// RequeuingStrategy defines the strategy for requeuing an evicted Workload.
```

bundle/manifests/kueue.openshift.io_kueues.yaml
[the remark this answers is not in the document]
MaysaMacedo: @kannon92 Doesn't the current Kueue upgrade process require uninstalling the kueue-operator?
would it require to also delete and create a new Kueue object CR? I wonder if that would give us some protection if the None field is gone. Of course we would need to document that very large timeouts would need to be set if the user wants to mimic a disable of the feature.

Another option is, we keep None as an option and kueue-operator sets really high timeouts on it to mimic disable of the feature.

pkg/apis/kueueoperator/v1/types.go
```diff
@@ -365,6 +363,82 @@ type ByWorkload struct {
...
+	// The current default is the value of timeoutSeconds.
+	// +kubebuilder:validation:Minimum=0
+	// +kubebuilder:validation:Maximum=31536000
+	// +optional
+	RecoveryTimeoutSeconds *int32 `json:"recoveryTimeoutSeconds,omitempty"`
+
+	// requeuingStrategy defines the strategy for requeuing a Workload that was evicted
+	// due to Pod readiness timeout.
+	// +optional
+	RequeuingStrategy *RequeuingStrategy `json:"requeuingStrategy,omitempty"`
```
everettraven: All the fields in the child type are optional, what is the difference between omitting this field entirely and setting `requeuingStrategy: {}`?

bundle/manifests/kueue.openshift.io_kueues.yaml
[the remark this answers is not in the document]
kannon92: We explictly don't require folks to remove the Kueue object CR during upgrade.

If we require the deletion of that object then we break the upgrade story.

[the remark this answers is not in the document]
kannon92: Removing Kueue object CR means that CQ/LQ would need to be removed. Which we want to avoid.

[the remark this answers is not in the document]
kannon92: We could decide to have 1.5 only work on 4.20+ but then we would support/maintain 1.4 until 4.18 and 4.19 are out of support.

pkg/apis/kueueoperator/v1/types.go
```diff
@@ -365,6 +363,82 @@ type ByWorkload struct {
...
+	TimeReference RequeuingTimeReference `json:"timeReference,omitempty"`
+
+	// backoffLimitCount defines the maximum number of re-queuing retries.
+	// Once the number is reached, the workload is deactivated (spec.active set to false).
+	// When omitted, kueue will decide the default, which is subject to change over time.
+	// The current default is null.
+	// +kubebuilder:validation:Minimum=1
+	// +kubebuilder:validation:Maximum=2000
+	// +optional
+	BackoffLimitCount int32 `json:"backoffLimitCount,omitempty"`
```
everettraven: Also not a huge deal since this matches the upstream configuration field name, but something like `RetryLimit` is more clear of a name IMO. The "Count" in the field name here just seems unnecessary.

pkg/apis/kueueoperator/v1/types.go
```diff
@@ -365,6 +363,82 @@ type ByWorkload struct {
...
+
+	// backoffBaseSeconds defines the base for the exponential backoff for re-queuing an evicted workload.
+	// The consecutive requeue delays are computed as backoffBaseSeconds * 2^(retryCount-1) + jitter.
+	// When omitted, kueue will decide the default, which is subject to change over time.
+	// The current default is 60 second.
+	// +kubebuilder:validation:Minimum=1
+	// +kubebuilder:validation:Maximum=3600
```
everettraven: How did we decide on these limits?

pkg/apis/kueueoperator/v1/types.go
```diff
@@ -365,6 +363,82 @@ type ByWorkload struct {
...
+	// +optional
+	BackoffBaseSeconds int32 `json:"backoffBaseSeconds,omitempty"`
+
+	// backoffMaxSeconds defines the maximum backoff time to re-queue an evicted workload.
+	// When omitted, kueue will decide the default, which is subject to change over time.
+	// The current default is 3600 seconds.
+	// +kubebuilder:validation:Minimum=1
+	// +kubebuilder:validation:Maximum=86400
```
everettraven: How did we decide on these limits?

pkg/apis/kueueoperator/v1/types.go
```diff
@@ -365,6 +363,82 @@ type ByWorkload struct {
...
+	// +optional
+	BackoffMaxSeconds int32 `json:"backoffMaxSeconds,omitempty"`
```
everettraven, line 447: What is the relationship between these two fields? Are there any constraints that should be enforced here?

For example, would something like:
```yaml
backoffBaseSeconds: 10
backoffMaxSeconds: 5
```
be considered a valid configuration?

bundle/manifests/kueue.openshift.io_kueues.yaml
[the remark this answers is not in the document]
MaysaMacedo: @kannon92 Just to make sure we are on the same page, the options to disable the feature either by setting really high timeouts or relying on the feature gate are not considered valid options, right?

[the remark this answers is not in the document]
kannon92: I think high timeouts is a valid option. 

I'd prefer to keep None and just set the values to disable it effectively.

pkg/apis/kueueoperator/v1/types.go
MaysaMacedo: Yes

pkg/apis/kueueoperator/v1/types.go
MaysaMacedo: The maximum (1 year) was defined based on conversation from upstream around allowing users to set really high value ,like 1 year to mimic the feature being disabled, for example https://github.com/kubernetes-sigs/kueue/pull/11855#discussion_r3511284407
I set the minimum to 1 sec to quickly be able to test an eviction. There are tests in upstream that even test to [miliseconds](https://github.com/kubernetes-sigs/kueue/blob/89dd4d718ef5e5b8127686181a2073205428645f/test/e2e/sequential/baseline/waitforpodsready_test.go#L84)
But I don't have a strong opinion on this.

pkg/apis/kueueoperator/v1/types.go
MaysaMacedo: Based on this comment https://github.com/openshift/kueue-operator/pull/2140#discussion_r3767814741

I will keep None as a valid option also in the Enum and set high timeouts to mimic a feature disable and will update the API docs. Does it make sense to you?

pkg/apis/kueueoperator/v1/types.go
MaysaMacedo: Maybe if we will enforce one year timeout when `None` is used on the Policy, we might not need to allow one year as max value here as well?
@kannon92 WDYT?

pkg/apis/kueueoperator/v1/types.go
MaysaMacedo: There is no difference between omitting and setting {}. Internally, when kueue-controller-manager sees a `requeuingStrategy` not set it will enforce the defaults values. https://github.com/kubernetes-sigs/kueue/blob/main/apis/config/v1beta2/defaults.go#L113-L116

pkg/apis/kueueoperator/v1/types.go
MaysaMacedo: Claude helped to define those values. 
`backoffBaseSeconds` is the base of the exponential function so it controls the floor of the wait after the first eviction. Setting 0 would mean no delay, that's why we have 1 as the minimum.
The maximum of the base to one hour might be too much, perhaps 5 minutes makes more  sense? this would allow more exponential retries to happen, also depending on the value of `BackoffMaxSeconds`. 
or maybe 10 minutes makes more sense? It really depends on the environment, for example workloads that are known to have slow start up or environments where the cluster recovers slowly would need bigger values on the base.
@kannon92 any thoughts on this?

pkg/apis/kueueoperator/v1/types.go
MaysaMacedo: No, it's not a valid configuration. We need: `BackoffBaseSeconds` <= `BackoffMaxSeconds` as it's the cap.

pkg/apis/kueueoperator/v1/types.go
kannon92: I think 1 year is fine.

pkg/apis/kueueoperator/v1/types.go
MaysaMacedo: Yes, it's equivalent. But I'm not sure if it's more accurate, this field controls:
- timeout to reach ready=true
- timeout to recover to reach ready=true
- what to do with the other workloads that are in the queue.
- when to requeue the workload.

pkg/apis/kueueoperator/v1/types.go
kannon92: I think this range is good for now. There wasn't much guidance on end-user thought with these fields so we are figuring it out as we go.

pkg/apis/kueueoperator/v1/types.go
everettraven: Seems fine to me.

pkg/apis/kueueoperator/v1/types.go
everettraven: Is 1 second actually a reasonable configuration that we would support? We try to avoid creating validations to enable easier testing and focus on what would actually be reasonable for end-users and what we would reasonably support.

pkg/apis/kueueoperator/v1/types.go
everettraven: If there is no difference, do not allow setting `{}` as a value and use omission of the field as a signal to default instead.

This can be done by either setting a minimum properties of 1 on the `RequeuingStrategy` type or making one of its fields required. This also makes it so you do not need to use a pointer.

pkg/apis/kueueoperator/v1/types.go
everettraven: Just a reminder that it is more acceptable (from the end-user perspective) to broaden the range in the future than it is to shrink it. I've got no strong opinions on the range, but wanted to make sure there was some reasonable backing behind the values chosen here - which I feel there is here.

pkg/apis/kueueoperator/v1/types.go
everettraven: Please add a validation that ensures that `backoffBaseSeconds` must be less than or equal to `backoffMaxSeconds`

pkg/apis/kueueoperator/v1/types.go
MaysaMacedo: Given it's more acceptable to broaden the range, I will update the backoffBaseSeconds minimum to ~~10~~ 30 seconds and keep the max of 1 hour.
The 10 seconds will allow fast recovery for transient issues .

pkg/apis/kueueoperator/v1/types.go
MaysaMacedo: I will update the `BackoffMaxSeconds` minimum to ~~10~~ 30 seconds to keep inline with `backoffBaseSeconds` and update  the max of 2 hours. I think it doesn't make sense to wait 24hrs between retries, I think the workload would be basically abandoned.

pkg/apis/kueueoperator/v1/types.go
MaysaMacedo: ~~2000 might be too aggressive, I will reduce it to 1000.~~
Even 1000 is too aggressive. Setting it to 200 based on the following:
- with the minimum settings if retries are 200 it would reach just under 2 hours
- with the max settings if retries are 200 it would reach around 2.5 weeks.
- with the default settings if retries are 200 it would reach around 8 days.
This calculation assumed each retry wouldn't take maximum timeout and the maximum recovery timeout.
Also the user has the option to not set it and the workload will be endless re-queued.

sohankunkerkar:
Putting this comment here since it applies to most, if not all, of the fields:

I don't think 1s would work. Pod creation, image pulling, and the readiness probe all happen before `PodsReady`, so we'd probably never hit that threshold. The Workload already records when it was admitted and when its pods became ready, so I think looking at those timestamps would give us a more realistic floor. 

For the maximums, I think it would be useful to spell out what each value actually costs. `backoffLimitCount` is especially interesting as a time-to-deactivation setting, since sufficiently large values effectively end up behaving the same as leaving it unset.

pkg/apis/kueueoperator/v1/types.go
MaysaMacedo: I ended up leaving as it given the field name already has time in it `timeReference`. Let me know if would prefer to have it updated and I will do it

MaysaMacedo:
@everettraven Thanks for the review! This PR is ready for another look when you can

pkg/apis/kueueoperator/v1/types.go
```diff
@@ -365,6 +365,84 @@ type ByWorkload struct {
...
+)
+
+// RequeuingStrategy defines the strategy for requeuing a Workload evicted.
+// +kubebuilder:validation:XValidation:rule="self.backoffBaseSeconds == 0 || self.backoffMaxSeconds == 0 || self.backoffBaseSeconds <= self.backoffMaxSeconds",message="backoffBaseSeconds must be less than or equal to backoffMaxSeconds"
+// +kubebuilder:validation:MinProperties=1
+type RequeuingStrategy struct {
+	// timeReference defines which point in time is used as the reference for re-queuing a Workload
+	// that was evicted. Allowed values are Eviction and Creation.
+	// Eviction uses the time of the Evicted condition with PodsReadyTimeout reason.
+	// Creation uses the Workload .metadata.creationTimestamp.
```
everettraven: Very minor nit (optional to address):

```suggestion
	// When set to 'Eviction', kueue will be configured to use the time of the Evicted condition with PodsReadyTimeout reason.
	// When set to 'Creation', kueue will be configured to use the Workload .metadata.creationTimestamp.
```

pkg/apis/kueueoperator/v1/types.go
```diff
@@ -365,6 +365,84 @@ type ByWorkload struct {
...
+	// When omitted, this means no opinion and the operator is left to choose a reasonable default, which is subject to change over time.
+	// The current default is Eviction.
+	// +optional
+	TimeReference RequeuingTimeReference `json:"timeReference,omitempty"`
+
+	// retryLimit defines the maximum number of re-queuing retries.
+	// Once the number is reached, the workload is deactivated (spec.active set to false).
+	// When omitted, this means no opinion and the operator is left to choose a reasonable default, which is subject to change over time.
+	// The current default is null, which means the workload will be requeued infinitely.
```
everettraven: Do we ever anticipate changing the default behavior here to set a default retry limit value?

everettraven (commented):
Overall, the structure looks good - one nit and one question

pkg/apis/kueueoperator/v1/types.go
MaysaMacedo: AFAIK, no.

MaysaMacedo: @kannon92 have you heard anything is this aspect?

kannon92: I haven't really heard much about these fields.

kannon92: I would prefer to let upstream decide these values as we don't have a lot of data to back up what makes sense here or not.

everettraven: I was more so asking to better understand if we need a way for an end-user to explicitly set "no limit" or if we are confident that omission of this configuration option will always mean "no limit". It is generally more intuitive for omission in this case to mean "no limit", but if we think that there is a chance that this will not always remain the case, we should allow an end user to specify some kind of explicit value that we use as a signal to not enforce a retry limit.

MaysaMacedo: The enforcement of the no-limit when null is done internally in the Kueue https://github.com/kubernetes-sigs/kueue/blob/main/pkg/controller/core/workload_controller.go#L1218-L1230
I don't see another way for us to enforce the no-limit given that is something done internally.

everettraven: Copying my question over from slack for posterity:

Do we have some level of trust that Kueue won't remove the ability for us to say "retry this thing infinitely" in one form or another?

pkg/apis/kueueoperator/v1/types.go
```diff
@@ -329,26 +329,26 @@ const (
...
 	// The allowed values are ByWorkload, None and "".
 	// When set to ByWorkload, this means each workload is processed and considered
 	// for admission as a single unit.
 	// Where workloads do not become ready over time, the entire workload may then be evicted and retried at a later time.
-	// None means gang scheduling is disabled.
+	// None means gang scheduling sets a really high admission timeout and does not block admission, which effectively disables gang scheduling.
 	// When set to "", this means no opinion and the operator is left
 	// to choose a reasonable default, which is subject to change over time.
-	// The current default is None.
+	// The current default is "".
```
everettraven: I understand that you are meaning for this to equate to "use kueue defaults".

You've already defined that "" means no opinion and that they are deferring to the operator to make a decision on their behalf. You cannot change the meaning of "" to anything other than that now.

Anything that you may default to also needs to be an explicit choice that the end-user can also set so they can specify an opinion of using that configuration and not be impacted by any defaulting behavior changes you make.

As a more concrete example, if you addsomething like `UseKueueDefaults` as an option, I can explicitly set `policy: UseKueueDefaults`. Even if you change the operator defaulting behavior to be `None` again on a future release, when I upgrade the operator should see that I specified that I have an opinion and respect my opinion instead of switching to `None`.

pkg/apis/kueueoperator/v1/types.go
MaysaMacedo: I can't say this for sure, but I find it really unlikely. Maybe @kannon92 or @sohankunkerkar has some idea.
In the meantime I'm updating the docs to:

> When omitted, no retry limit will be enforced and the workload will be re-queued infinitely until it has reached a ready state.

kannon92: TBH this feature hasn't been updated in a while and I imagine over time WAS would make some of these features redundant. `WaitForPodsReady` is mostly a safety hatch to not flood the queue with inadmission workloads.

pkg/configmap/configmap.go
```diff
@@ -139,12 +139,52 @@ func buildManagedJobsWithoutQueueName(workloadManagement kueue.WorkloadManagemen
...
+			w.RequeuingStrategy = buildRequeuingStrategy(&bw.RequeuingStrategy)
+		}
+		return w
+	case kueue.GangSchedulingPolicyNone:
+		return &configapi.WaitForPodsReady{
+			Timeout:        v1.Duration{Duration: 365 * 24 * time.Hour},
+			BlockAdmission: ptr.To(false),
+		}
 	default:
+		// Defaults to the values Kueue enforces internally for waitForPodsReady.
```
MaysaMacedo, line 162: @everettraven @kannon92 I would like you opinion on something.
the `ByWorkloadDefaults` and `""` options would end up in this case.
This means that `WaitForPodsReady: nil` would be used and the generated config-map won't have the option defined. However, the kueue-controller-manager enforces at runtime its defaults, which are:
`waitForPodsReady:\n  blockAdmission: false\n  recoveryTimeout: 30m0s\n  requeuingStrategy:\n    backoffBaseSeconds: 60\n    backoffMaxSeconds: 3600\n    timestamp: Eviction\n  timeout: 30m0s`
From a testing perspective we wouldn't be able to automate this by checking that the default values are really enforced, we would need to manually verify the controller logs.
Should I instead generate a  WaitForPodsReady with all the values we define as defaults in the API to give more security or we keep defaulting to whatever is in upstream?
Also note that I have changed the default value of the Policy to be `ByWorkloadDefaults`, let me know if you see any issues with that.

kannon92: I think we should focus on testing the behavior in our operator. I would prefer to just use the kueue defaults and not set them as they can change with response to the community.

So if `ByWorkloadsDefaults means `WaitForPodsReady: nil` is a fine test to add. And let kueue set the values accordingly. We don't need to test and/or verify the logs. Trust upstream to do the right thing here and we can enforce the test case in upstream if it doesn't.

pkg/apis/kueueoperator/v1/types.go
```diff
@@ -365,6 +368,84 @@ type ByWorkload struct {
...
+	// When set to 'Creation', kueue will be configured to use the Workload .metadata.creationTimestamp.
+	// When omitted, this means no opinion and the operator is left to choose a reasonable default, which is subject to change over time.
+	// The current default is Eviction.
+	// +optional
+	TimeReference RequeuingTimeReference `json:"timeReference,omitempty"`
+
+	// retryLimit defines the maximum number of re-queuing retries.
+	// Once the number is reached, the workload is deactivated (spec.active set to false).
+	// When omitted, no retry limit will be enforced and the workload will be re-queued infinitely until it has reached a ready state.
+	// The current default is null, which means the workload will be requeued infinitely.
```
everettraven:
```suggestion
```

pkg/configmap/configmap.go
everettraven: I agree with @kannon92 here.

everettraven:
Aside from my most recent comment, this LGTM from an API perspective.

PR #2151, opened 2026-07-06
update submodule july 6 main

kannon92:
## Summary by CodeRabbit

* **Bug Fixes**
  * Tightened validation so a workload’s queue assignment cannot be changed or removed while quota is still reserved.
  * Improved consistency checks to catch cases where the queue name is present on only one side of an update.


sohankunkerkar (approved):
/lgtm

PR #2282, opened 2026-07-30
OCPKUEUE-714: add TAS demo for quick examples

kannon92:
Walk through for TAS.

## Summary by CodeRabbit

* **New Features**
  * Added a complete Kubernetes Topology Aware Scheduling demo.
  * Added hostname- and datacenter-based scheduling scenarios with sample workloads, queues, resource configurations, and topology definitions.
  * Added examples for Jobs, Pods, JobSets, and Deployments using topology placement requirements and preferences.
  * Added scripts to configure and restore node labels for both scenarios.
* **Documentation**
  * Added step-by-step setup, verification, cleanup, and teardown instructions for the demos.

### DIFF
--- a/docs/tas/hostname/05-job.yaml
+++ b/docs/tas/hostname/05-job.yaml
@@ -1,0 +1,37 @@
+apiVersion: batch/v1
+kind: Job
+metadata:
+  generateName: tas-job-
+  namespace: kueue-tas-demo
+  labels:
+    kueue.x-k8s.io/queue-name: tas-hostname-lq
+spec:
+  # parallelism=2 with a required hostname topology means both pods must
+  # land on the *same* node (the only topology level here is the hostname
+  # itself). This is intentional: it demonstrates that TAS still admits and
+  # schedules the workload even on a single-worker-node cluster, and that
+  # the whole pod set is pinned to one topology domain.
+  parallelism: 2
+  completions: 2
+  template:
+    metadata:
+      annotations:
+        kueue.x-k8s.io/podset-required-topology: kubernetes.io/hostname
+    spec:
+      restartPolicy: Never
+      securityContext:
+        runAsNonRoot: true
+        seccompProfile:
+          type: RuntimeDefault
+      containers:
+      - name: test-container
+        image: registry.access.redhat.com/ubi9/ubi-minimal:latest
+        command: ["sh", "-c", "echo Hello Kueue; sleep 3600"]
+        resources:
+          requests:
+            cpu: 100m
+            memory: 64Mi
+        securityContext:
+          allowPrivilegeEscalation: false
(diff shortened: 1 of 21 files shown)


docs/tas/hostname/05-job.yaml
```diff
@@ -0,0 +1,32 @@
...
+  namespace: kueue-tas-demo
+  labels:
+    kueue.x-k8s.io/queue-name: tas-hostname-lq
+spec:
+  parallelism: 2
+  completions: 2
+  template:
+    metadata:
+      annotations:
+        kueue.x-k8s.io/podset-required-topology: kubernetes.io/hostname
```
sohankunkerkar, line 19: `required-topology: kubernetes.io/hostname` with parallelism=2 forces both pods onto the same node,is that the intent for this demo?

docs/tas/hostname/08-deployment.yaml
```diff
@@ -0,0 +1,31 @@
...
+  selector:
+    matchLabels:
+      app: test-deployment
+  template:
+    metadata:
+      labels:
+        app: test-deployment
+      annotations:
+        kueue.x-k8s.io/podset-required-topology: kubernetes.io/hostname
+    spec:
```
sohankunkerkar, line 22: Does this work in OCP without pod-level securityContext?

sohankunkerkar:
I have two questions, but overall, this is a pretty straightforward demo. Thanks for putting this together!

docs/tas/hostname/08-deployment.yaml
kannon92: Yes I think for high level controllers the get correctly defaulted.

kannon92: For pod integration that is where we need the securityContext. At least that has been my experience.

docs/tas/hostname/05-job.yaml
kannon92: yes I think that is what hostName would mean.

sohankunkerkar: Yeah, that's clear. Less clear is that the Job is the only one where the annotation does anything, so the Pod, both JobSet replicatedJobs, and the Deployment replicas are all single-pod groups, so "same node" is free. Could we add an inline comment like the datacenter ones have?

kannon92: Added a comment. thanks for the suggestion.

cpmeadors:
This looks really nice!  I would like to find time to actually run the demos before I give a real lgtm.

kannon92:
@sohankunkerkar 

can you take a quick look?

sohankunkerkar:
Thanks!
/lgtm

